PR #3, opened 2022-12-22
chore: store models on hf

qued:
Added code to download `oer-checkbox` model from a public [repo](https://huggingface.co/unstructuredio/oer-checkbox/) on huggingface. This should remove all dependencies on assets in private spaces and open the door for full use of this repo as a package.

Tested by building the container and deploying to a test EC2 instance.

#### Testing
```python
from unstructured_inference.models import get_model

get_model("checkbox")

```
Should show 2 downloading bars the first time you run this code, for the model binary and the model config.

Subsequent runs do not download anything (as long as the file is still present locally).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,6 @@
 ## 0.2.1
 
-* Update image link
+* Implement auto model downloading
 
 ## 0.2.0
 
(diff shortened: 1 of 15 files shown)


MthwRobinson (approved):
LGTM

PR #4, opened 2022-12-23
chore: logger should not be setting up a BasicConfig

yuming-long:
### Summary
Clean up loggers for `unstructured-inference` so it doesn't have BasicConfig, a logger named  `unstructured_inference` will be added to the [logger config file](https://github.com/Unstructured-IO/unstructured-infra-tools/blob/main/logger_config.yaml) in `unstructured-infra-tools`.

Related PR: https://github.com/Unstructured-IO/unstructured/pull/106

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,6 @@
-## 0.2.1
+## 0.2.1-dev0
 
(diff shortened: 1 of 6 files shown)


qued (approved):
LGTM!

PR #5, opened 2023-01-03
feat: Refactor for local inference

qued:
Small factorization, primarily moving some of the api processing code out to where it can be reused.

#### Testing:
Functionality should be unchanged from a user perspective.

Tests should pass.

Curl should work while app is running locally:
- Run `make run-app-dev`
- Run the following curls from the `pipeline-oer/sample-docs` directory, or replace `fake-oer.pdf` with the path to another pdf:
```shell
curl -X 'POST' \
  'localhost:8000/layout/pdf' \
  -H 'accept: application/json' \
  -H 'Content-Type: multipart/form-data' \
  -F 'file=@fake-oer.pdf' | jq -C . | less -R
```
Should succeed and return narrative sections.
```shell
curl -X 'POST' \
  'localhost:8000/layout/pdf' \
  -H 'accept: application/json' \
  -H 'Content-Type: multipart/form-data' \
  -F 'model=checkbox' \
  -F 'file=@fake-oer.pdf' | jq -C . | less -R
```
Should succeed and return checkboxes (should be present if `fake-oer.pdf` is used).
```shell
curl -X 'POST' \
  'localhost:8000/layout/pdf' \
  -H 'accept: application/json' \
  -H 'Content-Type: multipart/form-data' \
  -F 'model=fake' \
  -F 'file=@fake-oer.pdf' | jq -C . | less -R
```
Should fail with "unknown model" error message.

### DIFF
--- a/test_unstructured_inference/inference/test_layout.py
+++ b/test_unstructured_inference/inference/test_layout.py
@@ -157,6 +157,18 @@ def test_process_file_with_model(monkeypatch, mock_page_layout, model_name):
         "from_file",
         lambda *args, **kwargs: layout.DocumentLayout.from_pages([]),
     )
+    monkeypatch.setattr(
+        models, "load_model", lambda *args, **kwargs: MockLayoutModel(mock_page_layout)
+    )
+    monkeypatch.setattr(
+        models,
+        "_get_model_loading_info",
+        lambda *args, **kwargs: (
+            "fake-binary-path",
+            "fake-config-path",
+            {0: "Unchecked", 1: "Checked"},
+        ),
+    )
     filename = ""
     assert layout.process_file_with_model(filename, model_name=model_name)
 
--- a/test_unstructured_inference/inference/test_layout.py
+++ b/test_unstructured_inference/inference/test_layout.py
@@ -1,12 +1,14 @@
 import pytest
-from unittest.mock import patch
+from unittest.mock import patch, mock_open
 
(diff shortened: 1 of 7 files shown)


test_unstructured_inference/inference/test_layout.py
```diff
@@ -1,12 +1,14 @@
 import pytest
-from unittest.mock import patch
+from unittest.mock import patch, mock_open
```
MthwRobinson, line 2: Didn't know about `mock_open` before!

unstructured_inference/models/__init__.py
```diff
@@ -24,5 +24,11 @@ def _get_model_loading_info(model: str) -> Tuple[str, str, Dict[int, str]]:
         config_path = hf_hub_download(repo_id, config_fn)
         label_map = {0: "Unchecked", 1: "Checked"}
     else:
-        raise ValueError(f"Unknown model type: {model}")
+        raise UnknownModelException(f"Unknown model type: {model}")
     return model_path, config_path, label_map
+
+
+class UnknownModelException(Exception):
```
MthwRobinson, line 31: :-)

MthwRobinson (approved):
LGTM. I like `UnknownModelException` addition.

test_unstructured_inference/inference/test_layout.py
qued: Me either!

PR #6, opened 2023-01-04
chore: remove detectron from setup.py

qued:
Having a non-pip dependency prevents `unstructured-inference` from being installed as a package from pypi, so this PR removes detectron2 from the `setup.py` dependencies. Also removed the code to fail the install if `torch` isn't installed - `torch` will be installed with `layoutparser`, and will be present when the user builds Detectron2 from source.

Instructions have been updated to reflect the changes.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,4 @@
-## 0.2.1-dev1
+## 0.2.1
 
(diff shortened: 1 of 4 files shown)


MthwRobinson (approved):
LGTM

PR #7, opened 2023-01-06
chore: docker tweaks for repo consistency

cragwolfe:
Tweaks to `Dockerfile` / `make` commands to be more consistent (or soon to be consistent) with other Unstructured repos.


scripts/docker-build.sh
```diff
@@ -2,6 +2,7 @@
 
 set -euo pipefail
 
-DOCKER_BUILDKIT=1 docker buildx build --platform=linux/amd64 -f docker/Dockerfile \
+DOCKER_BUILDKIT=1 docker buildx build --platform=linux/amd64 -f Dockerfile \
+  --build-arg PIP_VERSION="$PIP_VERSION" \
   --progress plain \
-  -t ml-inference-dev:latest .
+  -t unstructured-inference-dev:latest .
```
qued, line 8: I left it ml-inference as far as the app/image name was concerned when I renamed the repo. Changing the name is fine, but references would need to be aligned in `deployment-tools` and `devops` (and maybe others that aren't coming to mind right now).

Makefile
```diff
@@ -50,14 +50,24 @@ pip-compile:
...
 
-## docker-build:            builds the docker container for detectron2
 .PHONY: docker-build
 docker-build:
-	PIP_VERSION=${PIP_VERSION}  ./scripts/docker-build.sh
+	PIP_VERSION=${PIP_VERSION} ./scripts/docker-build.sh
+
+.PHONY: docker-start-api
+docker-start-api:
+	docker run -p 8000:8000 --mount type=bind,source=$(realpath .),target=/home/notebook-user/local -t --rm --entrypoint uvicorn unstructured-inference-dev:latest ${PACKAGE_NAME}.api:app --host 0.0.0.0 --port 8000
```
qued, line 69:
```suggestion
	docker run -p 8000:8000 --mount type=bind,source=$(realpath .),target=/home/notebook-user/local -t --rm unstructured-inference-dev:latest --entrypoint uvicorn ${PACKAGE_NAME}.api:app --host 0.0.0.0 --port 8000
```
...although see my other note about changing the tag from `ml-inference`.

qued (approved):
LGTM!

Makefile
cragwolfe: actually, when using --entrypoint it needs to come before the image name. re-tested to make sure.
```
$ make docker-start-api
docker run -p 8000:8000 --mount type=bind,source=/home/ec2-user/repos/unstructured-inference,target=/home/\
notebook-user/local -t --rm --entrypoint uvicorn unstructured-inference-dev:latest unstructured_inference.\
api:app --host 0.0.0.0 --port 8000
INFO:     Started server process [1]
INFO:     Waiting for application startup.
INFO:     Application startup complete.
INFO:     Uvicorn running on http://0.0.0.0:8000 (Press CTRL+C to quit)
```

cragwolfe: (and need to use entrypoint since the Dockerfile defines it)

scripts/docker-build.sh
cragwolfe: discussed async. `unstructured-inference-dev` is at least consistent with other repos and should not lead to image-name confusion.

PR #8, opened 2023-01-08
feat: ocr when too many chars unrecognized

qued:
Certain PDF documents have text blocks within their layout that contain text, but the text contains / consists of unrecognized characters, represented in the text as `(cid:n)` where `n` is an integer ([ref](https://stackoverflow.com/questions/66656067/replace-cidnumber-with-chars-using-python-when-extracting-text-from-pdf-fil), [ref](https://stackoverflow.com/questions/50773909/what-to-do-with-cids-in-text-extracted-by-pdfminer)).

Current logic accepts whatever text is present, only applying OCR in the case the layout text is `None`. This PR changes the logic to also apply OCR in the case that > 50% of the characters are unrecognized.


unstructured_inference/inference/layout.py
```diff
@@ -156,3 +164,11 @@ def process_file_with_model(filename: str, model_name: str) -> DocumentLayout:
     model = None if model_name is None else get_model(model_name)
     layout = DocumentLayout.from_file(filename, model=model)
     return layout
+
+
+def cid_ratio(text: str) -> float:
+    """Gets ratio of unknown 'cid' characters extracted from text to all characters."""
```
cragwolfe, line 170: This feels not super cheap computationally, but I don't see a great way around it.
For minor optimization, could do:
```
if len(text) < 6:
   return 1.0
```

cragwolfe: actually, one could do plain old string.find for a first match of `(cid:` and return 1.0 if not found, otherwise the regex party.

MthwRobinson (commented):
Tested this on a document that was previously returning `cid`s and everything looks like it's working on this branch. Just one question from my end, I'll approve once you take a look at Crag's comments.

unstructured_inference/inference/layout.py
qued: Added some pre-checks before running regex as suggested.

cragwolfe (approved):

MthwRobinson (approved):
LGTM

unstructured_inference/inference/layout.py
```diff
@@ -156,3 +164,11 @@ def process_file_with_model(filename: str, model_name: str) -> DocumentLayout:
     model = None if model_name is None else get_model(model_name)
     layout = DocumentLayout.from_file(filename, model=model)
     return layout
+
+
+def cid_ratio(text: str) -> float:
```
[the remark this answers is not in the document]
qued: Looks like it's something `pdfminer.six` is doing: https://github.com/pdfminer/pdfminer.six/blob/20221105/pdfminer/converter.py#L235

PR #12, opened 2023-01-12
infra: Add PIP_VERSION ARG in dockerfile

stamzid:
Adding `PIP_VERSION` as a dockerfile ARG. Docker image has been built successfully locally with this update.

Reference: https://unstructured-ai.atlassian.net/browse/CORE-512


cragwolfe:
Let's update the Dockerfile to install pip3 in a manner consistent with our Makefile, e.g. https://github.com/Unstructured-IO/unstructured/blob/7b3b594/Makefile#L27 . Only, i think we'll want to use `pip3.8` to be consistent with other Dockerfiles that call `python3.8`.

cragwolfe (changes requested):
Forgot to request changes earlier, please see prior comment.

Dockerfile
```diff
@@ -6,6 +6,7 @@ FROM centos:centos7.9.2009
 #             https://mybinder.readthedocs.io/en/latest/tutorials/dockerfile.html
 ARG NB_USER=notebook-user
 ARG NB_UID=1000
+ARG PIP_VERSION=22.3.1
```
cragwolfe, line 9: I'd actually remove the default here since this is intended to always be explicitly [passed with scripts/docker-build.sh](https://github.com/Unstructured-IO/unstructured-inference/blob/main/scripts/docker-build.sh) .

cragwolfe (commented):
LGTM pending comment, confirmed `make docker-build` is install pip correctly.

Dockerfile
stamzid: sounds good

cragwolfe (approved):

PR #13, opened 2023-01-12
chore: cut release for image processing

qued:
Prepping release I realized some changes I thought I'd made never got in.
- Added `is_image` flag for `process_data_with_model` and `process_file_with_model`
- Added support for the above in the API
- Added testing for all of the above

#### Testing:

Run from python terminal:
```python

res = process_file_with_model('sample-docs/loremipsum.png', is_image=True)
res.pages[0].elements
print(res)

```
The output should show that several `Text` elements were found.

Run `make run-app-dev` and from a separate terminal make the following request:
```bash
curl -X 'POST' \
  'http://localhost:8000/layout/image' \
  -H 'accept: application/json' \
  -H 'Content-Type: multipart/form-data' \
  -F 'file=@sample-docs/loremipsum.png' | jq -C . | less -R

```
The output should show that several `Text` elements were found.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,4 @@
-## 0.2.2-dev1
+## 0.2.2
 
 * Add capability to process image files
 * Add logic to use OCR when layout text is full of unknown characters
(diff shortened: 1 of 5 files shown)


MthwRobinson (approved):
LGTM! Just note that I had to change the `process_file_with_model` line to the following in the test instructions for it to work:

```python
res = process_file_with_model(filename='sample-docs/loremipsum.png', model_name=None, is_image=True)
```

PR #19, opened 2023-01-13
chore: raise file not found for images

qued:
When trying to open an image in the `from_image_file` method, trying to open an empty filename would produce some non-helpful error, when ideally it should raise `FileNotFoundError`. This PR ensures that a `FileNotFoundError` is raised.

#### Testing:
With `main` branch active, run:
```python

DocumentLayout.from_image_file("")

```
Observe that you get an `AttributeError`. Run the code again from this branch, observe that you get a `FileNotFoundError`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,7 +1,6 @@
 ## 0.2.4-dev1
 
 * Clarify error when trying to open file that doesn't exist as an image
-* Requirements bumps (delete this line with first real CHANGELOG cahnge)
 
 ## 0.2.3
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,6 @@
-## 0.2.4-dev0
+## 0.2.4-dev1
 
+* Clarify error when trying to open file that doesn't exist as an image
 * Requirements bumps (delete this line with first real CHANGELOG cahnge)
 
 ## 0.2.3
--- a/test_unstructured_inference/inference/test_layout.py
+++ b/test_unstructured_inference/inference/test_layout.py
@@ -1,4 +1,5 @@
 import pytest
+import tempfile
 from unittest.mock import patch, mock_open
 
 import layoutparser as lp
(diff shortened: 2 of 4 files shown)


CHANGELOG.md
```diff
@@ -1,5 +1,6 @@
-## 0.2.4-dev0
+## 0.2.4-dev1
 
+* Clarify error when trying to open file that doesn't exist as an image
```
cragwolfe, line 3: please delete the next line CHANGELOG line :)

unstructured_inference/inference/layout.py
```diff
@@ -76,7 +77,13 @@ def from_file(cls, filename: str, model: Optional[Detectron2LayoutModel] = None)
     def from_image_file(cls, filename: str, model: Optional[Detectron2LayoutModel] = None):
         """Creates a DocumentLayout from an image file."""
         logger.info(f"Reading image file: {filename} ...")
-        image = Image.open(filename)
+        try:
+            image = Image.open(filename)
+        except Exception as e:
+            if os.path.isdir(filename) or os.path.isfile(filename):
```
MthwRobinson, line 83: If it's a directory, should we tell them it's a directory?

test_unstructured_inference/inference/test_layout.py
```diff
@@ -267,3 +267,8 @@ def mock_get_elements(self, *args, **kwargs):
         .elements
     )
     assert elements[0] == mock_page_layout
+
+
+def test_from_image_file_raises_with_empty_fn():
+    with pytest.raises(FileNotFoundError):
```
MthwRobinson, line 273: Should we also have a test for the case where the user passes in a directory?

MthwRobinson (approved):
Two minor/optional suggestions. LGTM once you take a look at those!

unstructured_inference/inference/layout.py
qued: From my testing passing a directory in this context gets handled properly by `Image.open`, raising an `IsADirectoryError`. I'll add a test case for this though.

test_unstructured_inference/inference/test_layout.py
qued: Case added in [ff0f5db](https://github.com/Unstructured-IO/unstructured-inference/pull/19/commits/ff0f5db55d0482be82f3ae71422bffde5bad163e).

PR #26, opened 2023-01-17
refactor: graceful exit if sed is an old version

gokullan:
This PR address [this issue](https://github.com/Unstructured-IO/community/issues/41) which asks for a graceful exit in cases where the version of sed is less than 4.3. I have made this version check in my commit and used exit 1 if the check fails. I edited the file inside GitHub itself. Hope it works as expected


cragwolfe (approved):
LGTM!

PR #38, opened 2023-01-26
refactor: generic model

qued:
Refactor of model code to make it less dependent on detectron2. 

Creates an abstract model class `UnstructuredModel`, with inference triggered by calling the object. The class must be subclassed for each individual model type to at least guide the inference input to the right method (e.g. the `detect` method for detectron 2 models).

Detectron specific code has been moved to `unstructured_inference.models.detectron2`.

The code still uses layoutparser objects for convenience in the `layout` code preparing the input, and the output is assumed to be a layoutparser `Layout` object. We may not want output in this format for other model types, but it seems like a viable choice for an output format, so we might consider wrangling the format to this class.

#### Testing:

Tests should be passing and the high level interface should operate as usual.

### DIFF
--- a/test_unstructured_inference/models/test_detectron2.py
+++ b/test_unstructured_inference/models/test_detectron2.py
@@ -11,7 +11,11 @@ class MockDetectron2LayoutModel:
         self.kwargs = kwargs
 
     def detect(self, x):
-        return "called"
+        return MockLayout()
+
+
+class MockLayout:
+    pass
 
 
 def test_load_default_model(monkeypatch):
@@ -40,4 +44,4 @@ def test_load_model(monkeypatch, config_path, model_path):
 
 def test_unstructured_detectron_model():
     model = detectron2.UnstructuredDetectronModel(MockDetectron2LayoutModel())
-    assert model(None) == "called"
+    assert isinstance(model(None), MockLayout)
--- a/test_unstructured_inference/models/test_detectron2.py
+++ b/test_unstructured_inference/models/test_detectron2.py
@@ -17,7 +17,7 @@ def test_load_default_model(monkeypatch):
     with patch.object(detectron2, "is_detectron2_available", return_value=True):
         model = models.get_model()
 
-    assert isinstance(model, MockDetectron2LayoutModel)
+    assert isinstance(model.model, MockDetectron2LayoutModel)
 
 
 def test_load_default_model_raises_when_not_available():
@@ -31,5 +31,5 @@ def test_load_model(monkeypatch, config_path, model_path):
     monkeypatch.setattr(detectron2, "Detectron2LayoutModel", MockDetectron2LayoutModel)
     with patch.object(detectron2, "is_detectron2_available", return_value=True):
         model = detectron2.load_model(config_path=config_path, model_path=model_path)
-    assert config_path == model.args[0]
-    assert model_path == model.kwargs["model_path"]
+    assert config_path == model.model.args[0]
+    assert model_path == model.model.kwargs["model_path"]
--- a/test_unstructured_inference/models/test_detectron2.py
+++ b/test_unstructured_inference/models/test_detectron2.py
@@ -10,6 +10,9 @@ class MockDetectron2LayoutModel:
         self.args = args
         self.kwargs = kwargs
 
+    def detect(self, x):
+        return "called"
+
 
 def test_load_default_model(monkeypatch):
     monkeypatch.setattr(detectron2, "Detectron2LayoutModel", MockDetectron2LayoutModel)
@@ -33,3 +36,8 @@ def test_load_model(monkeypatch, config_path, model_path):
         model = detectron2.load_model(config_path=config_path, model_path=model_path)
     assert config_path == model.model.args[0]
     assert model_path == model.model.kwargs["model_path"]
+
+
+def test_unstructured_detectron_model():
+    model = detectron2.UnstructuredDetectronModel(MockDetectron2LayoutModel())
+    assert model(None) == "called"
--- a/test_unstructured_inference/models/test_detectron2.py
+++ b/test_unstructured_inference/models/test_detectron2.py
@@ -37,11 +37,13 @@ def test_load_default_model_raises_when_not_available():
 def test_load_model(monkeypatch, config_path, model_path):
     monkeypatch.setattr(detectron2, "Detectron2LayoutModel", MockDetectron2LayoutModel)
     with patch.object(detectron2, "is_detectron2_available", return_value=True):
-        model = detectron2.load_model(config_path=config_path, model_path=model_path)
+        model = detectron2.UnstructuredDetectronModel()
+        model.initialize(config_path=config_path, model_path=model_path)
     assert config_path == model.model.args[0]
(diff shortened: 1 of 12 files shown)


test_unstructured_inference/models/test_detectron2.py
```diff
@@ -10,14 +10,17 @@ def __init__(self, *args, **kwargs):
         self.args = args
         self.kwargs = kwargs
 
+    def detect(self, x):
+        return "called"
```
MthwRobinson, line 14: Is this supposed to return a string? Totally fine if so, just felt a little odd.

test_unstructured_inference/test_utils.py
```diff
@@ -0,0 +1,44 @@
...
+    assert all(kd == kld for kd, kld in zip(d, ld))
+    assert all(d[k] == ld[k] for k in d)
+    assert len(ld) == len(d)
+
+
+def test_lazy_evaluate():
+    called = 0
+
+    def func(x):
+        nonlocal called
```
MthwRobinson, line 17: Interesting. I knew about `global` before but didn't realize there was a `nonlocal` too

unstructured_inference/models/detectron2.py
```diff
@@ -21,6 +24,43 @@
 DEFAULT_EXTRA_CONFIG: Final[List[Any]] = ["MODEL.ROI_HEADS.SCORE_THRESH_TEST", 0.8]
 
 
+# NOTE(alan): Entries are implemented as LazyDicts so that models aren't downloaded until they are
```
MthwRobinson, line 27: Is there a supported way to download the models ahead of time if you want to do it ahead of time? For instance if you want to include the model in a docker image or something like that?

MthwRobinson (commented):
Implementation looks good, just a couple of questions

test_unstructured_inference/test_utils.py
qued: Yeah I had trouble remember how to get the scope one level up and had to look it up.

unstructured_inference/models/detectron2.py
qued: Nothing out of the box. We could do it by running some code that instantiates whichever model we want, or we could build something to do it.

MthwRobinson: As long as you can do something along the lines of what we do for NLTK, we're probably fine 

https://github.com/Unstructured-IO/unstructured/blob/main/Makefile#L37-L38

test_unstructured_inference/models/test_detectron2.py
qued: I don't do anything with the output. Maybe it's a bit misleading this way though.

MthwRobinson: Yeah I'd probably return `None` if you don't do anything with it, but I don't feel too strongly about it

MthwRobinson (approved):
LGTM!

unstructured_inference/models/unstructuredmodel.py
```diff
@@ -0,0 +1,14 @@
+from abc import ABC, abstractmethod
+from typing import Any
+from PIL import Image
+
+
+class UnstructuredModel(ABC):
+    """Wrapper class for the various models used by unstructured."""
+
+    def __init__(self, model: Any):
```
cragwolfe, line 9: What is model supposed to be, a string, an object, either?
Could use some docstring here.

unstructured_inference/models/unstructuredmodel.py
```diff
@@ -0,0 +1,14 @@
...
+        self.model = model
+
+    @abstractmethod
+    def __call__(self, x: Image) -> Any:
```
cragwolfe, line 13: If we only accept Images, should this be called `UnstructuredImageModel`?

cragwolfe: not sure how i feel about objects that have the `__call__` method overridden, can make for hard to read code.
Is there any reason why the interface can't be `infer(...)`

unstructured_inference/models/unstructuredmodel.py
cragwolfe: Can this abstract class not bookkeep the model?

Use triton as an example, there interface methods are:

```
   initialize(self, ) [can optionally pass in model_name, config, version, etc]
   execute(self, requests): do an inference for each request
   finalize(self, ): cleanup, "unload"
   auto_complete_config(): the default config
```
   
 Reference:
 https://github.com/triton-inference-server/python_backend/blob/main/examples/pytorch/model.py
 https://github.com/triton-inference-server/python_backend

cragwolfe: On __init__ vs. initialize. If there is going to be any real work involved, it probably belongs in `initialize()`. I.e., it should be possible to instantiate a class without bogging down your system to do it.

unstructured_inference/models/unstructuredmodel.py
qued: I guess there's no harm in making it more general here. Changed input type to `Any` and to `Image` in the detectron2 subclass.

unstructured_inference/models/unstructuredmodel.py
qued: Added a docstring.

test_unstructured_inference/models/test_detectron2.py
qued: Changed output type to `MockLayout`

cragwolfe (approved):
Approving the PR so as to not block, but curious thoughts on other comments.

unstructured_inference/models/unstructuredmodel.py
qued: For now, the only work to be done is to download the binary and instantiate the model being wrapped, all of of which is already done by the time this wrapper model is instantiated (since you have to pass it the model binary). We could do it a different way, and include the `load_model` logic in the class itself.

unstructured_inference/models/unstructuredmodel.py
MthwRobinson: The `__call__` pattern for inference is what `transformers` does for `pipelines`. I actually don't mind that pattern here because it would be familiar to most of the people using this code.

qued: IMO a good pattern/name here would be something commonly used by other ML libraries for model inference.

I wouldn't be surprised if `infer` is used in an ML library, but I'm not aware of any.

I went with `__call__` since model objects you call for inference is a well established pattern found in torch and tensorflow/keras and propagates to libraries that wrap those such as fastai (although in each of these there are alternative ways to call the model for inference, `forward` in torch, `predict` in keras). Other common ones that I'm aware of are `forward` (didn't really like this because it suggests a feedforward process that's incidental to what we're doing here), `predict`/`predict_proba` (from `sklearn`), and most recently to me, `detect` from detectron. Of these I like `predict` the best. Maybe we could allow both `predict` and `__call__`.

MthwRobinson (approved):
LGTM. Re-reviewing after the addition of `predict` and the update to `__call__`.

PR #49, opened 2023-02-09
test: Reduced time for testing

benjats07:
Added some @pytest.mark.slow decorators to reduce the time of execution of the tests in CI, tests marked with it uses a tiny version of the YoloX model is loaded.

To run tests just need to run `make test` as usual and the models will not be tested on full capabilities, with `make test-slow` the test is fully executed.

In this PR, was improved the processing time when OCR and YoloX are used. Now the elements detected on the document are processed concurrently using multiprocessing.Pool.


test_unstructured_inference/models/test_yolox.py
```diff
@@ -0,0 +1,213 @@
...
+import pytest
+from fastapi.testclient import TestClient
+
+from unstructured_inference import api
+from unstructured_inference.inference.layout import DocumentLayout
+from unstructured_inference.models.yolox import yolox_local_inference, _get_model_loading_info
+from unstructured_inference.models import UnknownModelException
+
+
+@pytest.mark.skipif(
```
qued, line 14: Pytest has a pattern that I think fits better here. If you decorate the long-running tests with something like `@pytest.mark.long`, then in the make target for only short running tests you can run `pytest -m "not long" test_${PACKAGE_NAME} --cov=${PACKAGE_NAME} --cov-report term-missing`. See [the pytest markers documentation](https://docs.pytest.org/en/latest/example/markers.html) for more examples and explanation.

benjats07: Just changed it to that way, it's much cleaner. Thanks!

test_unstructured_inference/models/test_yolox.py
```diff
@@ -0,0 +1,195 @@
...
+    # NOTE(benjamin) The example sent to the test contains 5 detections
+    assert len(document_layout.pages[0].layout) == 0
+
+
+def test_invalid_model():
+    with pytest.raises(UnknownModelException):
+        get_model_loading_info("invalidmodel")
+
+
+# Only short test below
```
MthwRobinson, line 110: `test` -> `tests`. I'd also make this more prominent, like:
```python
######################
ONLY SHORT TESTS BELOW
######################
```

test_unstructured_inference/test_api.py
```diff
@@ -53,89 +50,10 @@ def test_layout_parsing_api(monkeypatch, filetype, ext, data, process_func, expe
 def test_bad_route_404():
     client = TestClient(api.app)
     filename = os.path.join("sample-docs", "loremipsum.pdf")
-    response = client.post("/layout/badroute", files={"file": (filename, open(filename, "rb"))})
-    assert response.status_code == 404
-
-
-def test_layout_v02_api_parsing_image():
```
MthwRobinson, line 60: Looks like a lot of these tests were moved to `test_yolox.py`. Did you make changes to these tests or did they just move?

unstructured_inference/inference/layout.py
```diff
@@ -142,28 +143,44 @@ def parse_image_elements(self, filename, num, DPI=500):
...
-                cv2.imwrite(dest_file, patch)
-                # Enabling this makes test_load_agent fails
-                if not tesseract.ocr_agent:
-                    tesseract.load_agent()
-                text = tesseract.ocr_agent.detect(patch)
-
-                element.text = text
-                new_layout.append(element)
-
+            pool = Pool()
```
MthwRobinson, line 146: What does `pool` do?

unstructured_inference/inference/layout.py
```diff
@@ -142,28 +143,44 @@ def parse_image_elements(self, filename, num, DPI=500):
...
+        upper_left_y = int(upper_left_y)
+        width = upper_left_x + int(element.get_width())
+        height = upper_left_y + int(element.get_height())
+        dest_file = os.path.join(tmp_folder, f"{n_page}-{n_element}.jpg")
+
+        image = cv2.imread(filename)
+        patch = image[upper_left_y:height, upper_left_x:width]
+        try:
+            cv2.imwrite(dest_file, patch)
+        except Exception as E:
```
MthwRobinson, line 171: Is there a specific exception type you'd expect there? Would be better to have this catch a specific error type. Also minor nit, `E` -> `e`.

unstructured_inference/models/yolox.py
```diff
@@ -13,14 +13,42 @@
...
 from unstructured_inference.inference.layout import DocumentLayout, LayoutElement, PageLayout
-from unstructured_inference.models import _get_model_loading_info
 from unstructured_inference.visualize import draw_bounding_boxes
+from unstructured_inference.models.base import UnknownModelException
+import json
+from typing import Tuple, Dict
+from huggingface_hub import hf_hub_download
+
+
+def get_model_loading_info(model: str) -> Tuple[str, Dict[int, str]]:
```
MthwRobinson, line 23: Is this new?

MthwRobinson (changes requested):
Added some inline comments and minor suggestions. Also, looks like there were some application code changes in addition to the test updates. Could you describe what those changes are in the PR description and add instructions for testing?

test_unstructured_inference/test_api.py
benjats07: Just moved to that file

unstructured_inference/inference/layout.py
benjats07: Is an object to create multiple concurrent processes, in this case for concurrent OCR on images with tesseract

unstructured_inference/models/yolox.py
benjats07: Used to have the _get_model_loading_info name in __init__.py, @qued / @cragwolfe mentioned that that file should keep empty

unstructured_inference/inference/layout.py
MthwRobinson: 👍 - Could you add a comment about that above this block?

MthwRobinson (approved):
Comments addressed, LGTM!

Issue #51, opened 2023-02-10
from_file broken, len(images) != len(layout)


hyperknot:
The from_file method in inference.layout.DocumentLayout.from_file is broken.

First, the load_pdf is duplicated 2x, which I guess it probably the hot-path.

Second, the loop breaks for me because len(images) != len(layout). I mean we are looping through layouts and then loading the image based on the layout's index? But for me the images != layouts. Actually len(images) = 0, is this because I have some dependency missing which is failing silently?

Are len(images) supposed to equal len(layout)? I'm running it on the official loremipsum.pdf as well. Platform is macOS.

```
layout = DocumentLayout.from_file("sample-docs/loremipsum.pdf")
```

https://github.com/Unstructured-IO/unstructured-inference/blob/6d2205b3e6fc6aa9b5bc7f4a1839fd66fd40413e/unstructured_inference/inference/layout.py#L74-L91

hyperknot:
The solution was to `brew install poppler`. It's a pity that pdf2image and thus layoutparser silently fails if poppler is not installed. Probably a check has to be added in from_file which checks if the len()-s are not equal and raises an exception to please install Poppler.

The duplicated hot path is still a valid bug though.

qued:
Thanks for finding this. I'm doing a [quick update](https://github.com/Unstructured-IO/unstructured-inference/pull/52) to remove the duplicate line, then I'll try to reproduce the mismatched `len` behavior.

PR #53, opened 2023-02-10
fix: catch len mismatch

qued:
From [issue 51](https://github.com/Unstructured-IO/unstructured-inference/issues/51) we have a report of `load_pdf` coming back with a non-empty list of layouts and an empty list of images. The reported fix was to install poppler.

I haven't been able to reproduce yet (if I try `from_file` without poppler installed I get a `RuntimeError` with a suggestion to install poppler), but it doesn't hurt to guard against a length mismatch, as that shouldn't happen and would cause subsequent code to fail. I added an error message that points the user at a possible poppler install issue.

#### Testing:
I added a test that checks the error raises properly, so just look over the code and make sure it makes sense.

### DIFF
--- a/unstructured_inference/inference/layout.py
+++ b/unstructured_inference/inference/layout.py
@@ -81,7 +81,7 @@ class DocumentLayout:
         layouts, images = load_pdf(filename, load_images=True)
         if len(layouts) > len(images):
             raise RuntimeError(
-                "Some images were not loaded. Check that poppler is installed and in PATH."
+                "Some images were not loaded. Check that poppler is installed and in your $PATH."
             )
         pages: List[PageLayout] = list()
         for i, layout in enumerate(layouts):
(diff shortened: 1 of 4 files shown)


unstructured_inference/inference/layout.py
```diff
@@ -79,6 +79,10 @@ def from_file(cls, filename: str, model: Optional[Detectron2LayoutModel] = None)
         # image and returns a dict, or something.
         logger.info(f"Reading PDF for file: {filename} ...")
         layouts, images = load_pdf(filename, load_images=True)
+        if len(layouts) > len(images):
+            raise RuntimeError(
+                "Some images were not loaded. Check that poppler is installed and in PATH."
```
cragwolfe, line 84: "in your $PATH" maybe?

cragwolfe (approved):

MthwRobinson (approved):
LGTM. Good catch!

PR #56, opened 2023-02-21
refactor: layout and model

qued:
Updated so that YoloX uses the same code paths as other models.
- Removed code that was duplicating existing functions
- Integrated YoloX model into existing pipeline
  - This means YoloX now has an associated `UnstructuredModel` called `UnstructuredYoloXModel`
  - The output of YoloX inference is now a `layoutparser` `Layout` object, like the other `UnstructuredModel`s so that it can integrate with the `DocumentLayout`, `PageLayout`, and `LayoutElement` classes
  - Other functionality is provided by the `DocumentLayout`, `PageLayout`, and `LayoutElement` classes
- Removed `jsons` dependency that was used in testing and updated deps. The dependency had a cool feature that allowed it to construct non-json-serializable classes from a json, but since that was being used to check content, content could be checked through another method.
- One piece of functionality that was in the code prior to this change and isn't now is the ability to create test artifacts that can be viewed for debugging purposes. This can be added back in, but artifacts must be created from the new outputs of the model (either a `layoutparser` `Layout` object or `PageLayout` object).

This PR also allows the capability to create a `PageLayout` with a fixed layout, i.e. the layout and location of text elements is specified at construction rather than inferred by model. This is useful for use cases where the layout of a particular form is known.


CHANGELOG.md
```diff
@@ -1,5 +1,6 @@
-## 0.2.8-dev0
+## 0.2.8-dev2
```
MthwRobinson, line 1: Do we want to do a release with this PR?

test_unstructured_inference/inference/test_layout.py
```diff
@@ -84,20 +84,27 @@ def test_get_page_elements(monkeypatch, mock_page_layout):
     assert elements == page.elements
 
 
+class MockPool:
```
MthwRobinson, line 87: 🐟

MthwRobinson (approved):
LGTM, nice refactor! Based on this, are are we close to being able to swap out detectron for yolox in `unstructured`?

unstructured_inference/models/yolox.py
```diff
@@ -2,208 +2,129 @@
...
+class UnstructuredYoloXModel(UnstructuredModel):
+    def predict(self, x: Image):
+        super().predict(x)
+        return self.image_processing(x)
+
+    def initialize(self, model_path: str, label_map: dict):
+        self.model = model_path
+        self.layout_classes = label_map
+
+    def image_processing(
```
benjats07, line 57: The deleted image_processing allows reusing the onnxruntime.InferenceSession (receives it as an argument), would be great to recover that for avoiding reloading the model.

unstructured_inference/api.py
```diff
@@ -2,28 +2,31 @@
...
-from unstructured_inference.models.yolox import yolox_local_inference
 
 app = FastAPI()
 
 ALL_ELEMS = "_ALL"
 VALID_FILETYPES = ["pdf", "image"]
 
 
-@app.post("/layout/detectron/{filetype:path}")
+@app.post("/layout/{modeltype:path}/{filetype:path}")
```
benjats07, line 12: Amazing, everything in just one API definition 🤯

benjats07 (approved):
In general LGTM.
I can try to recover the drawing of the results in the future (and this time, thanks to the changes will cover both models!) however, this PR seems a very good refactor 😄

CHANGELOG.md
qued: Sure, I'll update.

unstructured_inference/models/yolox.py
qued: Ah, thanks for flagging this.

qued: I made some changes in [620550f](https://github.com/Unstructured-IO/unstructured-inference/pull/56/commits/620550f0171349e7ab21ee17f6e59c90cc334c01) to persist the inference session in the model object. See if that addresses your concern here.

benjats07: Yes, I was thinking something like that.

PR #65, opened 2023-03-02
fix: removes multithreading for OCR

benjats07:
The multithreading problem doesn't appears to happen on M1 chip (as it does in x64 processors)
This PR just removes Pool from DocumentLayout.get_elements_from_layout.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,7 @@
+## 0.2.9
+
+* Removed multithreading from OCR
+
 ## 0.2.8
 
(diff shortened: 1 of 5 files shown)


ajjimeno:
It restarted the checks for [5460b58](https://github.com/Unstructured-IO/unstructured-inference/pull/65/commits/5460b58a9302eb7808ffb00c7da45b12d60d4b48) but there is one more commit for which I cannot start the CI tests. Probably the last commit needs to be ok to go ahead.

ajjimeno (approved):
LGTM

PR #68, opened 2023-03-09
fix: ocr when no elements are found in block

qued:
Stopgap fix for a bug that causes the parsing procedure to ignore pdf elements that are not contained within the bounds of an inferred/specified layout element.

#### Testing:
With the following linked file, [loremipsum-flat.pdf](https://github.com/Unstructured-IO/unstructured-inference/files/10927802/loremipsum-flat.pdf), try the code:
```python
doc = DocumentLayout.from_file('loremipsum-flat.pdf')
print(' '.join([el.text for el in doc.pages[0].elements]))

```
Observe that the output is blank on the main branch, but contains text on this branch.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,7 @@
+## 0.2.11
+
+* Fixed some cases where image elements were not being OCR'd
+
 ## 0.2.10
 
 * Removed control characters from tesseract output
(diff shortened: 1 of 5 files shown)


benjats07 (approved):
LGTM

PR #69, opened 2023-03-09
Fix: RGBA images

benjats07:
This PR fixes a problem with images with RGBA format, since them have 4 channels instead of expected 3, inference in models (YoloX, at least) fails.

Just needed add convert("RGB") to fix this problem, the corresponding test is added together with an example in sample-docs/RGBA_image.png 

```
curl -X 'POST' 'http://localhost:8000/layout/yolox/pdf' \
-F 'file=@sample-docs/RGBA_image.png' \
-F force_ocr=true 
 | jq -C | less -R
```


qued (approved):
LGTM!

PR #70, opened 2023-03-10
Integrated PyMuPDF

HAKSOAT:
Closes [310](https://github.com/Unstructured-IO/unstructured/issues/310)

Makes use of PyMuPDF to extract images in cases where `layoutparser.io.pdf`'s `load_pdf` doesn't extract images as desired.


HAKSOAT:
Hello @MthwRobinson I'd appreciate a review on this. Thank you.

cragwolfe:
Perhaps at the module level in`unstructured_inference/inference/layout.py` a warning could be emitted if `poppler` is not available to avoid the case where users are getting potentially lower quality outputs without realizing it (assuming that is the case). Ie, only emit the warning at import time rather than per document.

Actually, could take it one step further if neither are available and also emit a warning for that.

HAKSOAT:
> Perhaps at the module level in`unstructured_inference/inference/layout.py` a warning could be emitted if `poppler` is not available to avoid the case where users are getting potentially lower quality outputs without realizing it (assuming that is the case). Ie, only emit the warning at import time rather than per document.
> 
> Actually, could take it one step further if neither are available and also emit a warning for that.

Makes sense. I'll work on it.

unstructured_inference/inference/layout.py
```diff
@@ -89,9 +91,20 @@ def from_file(
...
+                    pixmap = page.get_pixmap()
+                    images.append(Image.open(io.BytesIO(pixmap.pil_tobytes("ppm"))))
+                if len(layouts) > len(images):
+                    raise ValueError("PyMuPDF extracted less images than layouts.")
+
+            except Exception as exc:
+                raise RuntimeError(
+                    "Some images were not loaded. "
+                    "Check that poppler is installed and in your $PATH."
+                ) from exc
```
HAKSOAT, line 106: Hi @cragwolfe I took a look at the suggestion. I think an import-level warning may not actually do the job we desire here since we are actually raising an error here. Hence, the extraction process is expected to end regardless of whether we have a warning or not.

cragwolfe:
Closing given license concern.

PR #71, opened 2023-03-22
feat: better image support

qued:
Updated the OCR logic to be aware of image elements.

LayoutParser only deals with text objects, so this PR removes LayoutParser from the internals and replaces the components. (LayoutParser remains a dependency because of detectron2.)

#### Testing:
Run:
```python

doc = DocumentLayout.from_file('sample-docs/loremipsum-flat.pdf')

```
`doc.pages[0].elements` should contain elements with the text of the document.


mallorih (approved):
LGTM

benjats07 (approved):
LGTM

PR #75, opened 2023-03-30
fix: make paddleocr install optional

qued:
Made `paddleocr` optional again, to avoid issues with `unstructured` docker image until we can resolve how to properly install `paddleocr` in Centos7.


MthwRobinson (approved):
LGTM! Looks optional to me

PR #76, opened 2023-03-31
feat: extend extract tables through interface

qued:
Allows `extract_tables` parameter to be passed from higher level functions `process_file_with_model` and `process_data_with_model`.

#### Testing:
Run:
```python
process_file_with_model("sample-docs/loremipsum.pdf", model_name="yolox", extract_tables=False)
```
Should generate no errors.


benjats07 (approved):
LGTM

PR #79, opened 2023-04-03
build: update detectron2 version

qued:
Update `detectron2` version to be consistent across repos.

#### Testing:

CI should pass.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -46,7 +46,7 @@ COPY unstructured_inference unstructured_inference
 RUN python3.8 -m pip install pip==${PIP_VERSION} \
   && pip3.8 install  --no-cache  -r requirements-base.txt \
   && pip3.8 install  --no-cache  -r requirements-dev.txt \
-  && pip3.8 install --no-cache "detectron2@git+https://github.com/facebookresearch/detectron2.git@v0.6#egg=detectron2" \
+  && pip3.8 install --no-cache "detectron2@git+https://github.com/facebookresearch/detectron2.git@e2ce8dc1ab097891395d324abaffe9cf298503d1#egg=detectron2" \
   && python3.8 -c "import unstructured_inference.models.detectron2 as detectron2; detectron2.load_default_model()"
 
 EXPOSE 5000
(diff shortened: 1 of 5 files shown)


Makefile
```diff
@@ -29,7 +29,7 @@ install-base-pip-packages:
 
 .PHONY: install-detectron2
 install-detectron2:
-	pip install "detectron2@git+https://github.com/facebookresearch/detectron2.git@78d5b4f335005091fe0364ce4775d711ec93566e"
+	pip install "detectron2@git+https://github.com/facebookresearch/detectron2.git@e2ce8dc1ab097891395d324abaffe9cf298503d1#egg=detectron2"
```
benjats07, line 32: LGTM
Just curiosity: there is any reason to choose that specific commit of detectron2 repo?

benjats07 (approved):

Makefile
qued: I copied it from what's in the `Makefile` of the `unstructured` repo, and if I had to guess why it was chosen, I'd say it was probably the most recent version with passing tests at the time it was chosen a couple of weeks ago. Let me know if there's a concern about using this version vs the one you had.

benjats07: Yeah, I think the same, no problem.

Dockerfile
```diff
@@ -46,7 +46,7 @@ COPY unstructured_inference unstructured_inference
 RUN python3.8 -m pip install pip==${PIP_VERSION} \
   && pip3.8 install  --no-cache  -r requirements-base.txt \
   && pip3.8 install  --no-cache  -r requirements-dev.txt \
-  && pip3.8 install --no-cache "detectron2@git+https://github.com/facebookresearch/detectron2.git@v0.6#egg=detectron2" \
+  && pip3.8 install --no-cache "detectron2@git+https://github.com/facebookresearch/detectron2.git@e2ce8dc1ab097891395d324abaffe9cf298503d1#egg=detectron2" \
```
cragwolfe, line 49: actually, since we're slinging this pip install all over the place, can we truncate the hash, e.g.:

```
pip3.8 install --no-cache "detectron2@git+https://github.com/facebookresearch/detectron2.git@e2ce8dc#egg=detectron2"
```

qued: Done in [17534b2](https://github.com/Unstructured-IO/unstructured-inference/pull/79/commits/17534b299359535c1164adce7bf8e1a1027b1101).

PR #80, opened 2023-04-09
fix: image location

qued:
Fixed bug where wrong location keys were being used to vertically locate image objects, causing the images' vertical coordinates to be given in terms of distance from the bottom of the document while text elements had vertical coordinates in terms of distance from the top of the document.

Added test that fails before change, passes after change.

#### Testing:

Ensure test `test_load_pdf_image_placement` fails on `main` branch, passes CI.


benjats07 (approved):
LGTM

PR #98, opened 2023-05-04
fix: patch pdfminer

qued:
Addresses [#346](https://github.com/Unstructured-IO/unstructured/issues/346) (need to push changes to `unstructured` to close).

`pdfminer.six` is not parsing keywords if the end of the keyword coincides with the end of a stream of PDF operations. I've created a [PR](https://github.com/pdfminer/pdfminer.six/pull/885) into `pdfminer.six` to address this, but while that is working its way through, I've created a patch here.

#### Testing:
Processing `sample-docs/IRS-form-1987.pdf` should no longer produce empty results:
```python

doc = DocumentLayout.from_file("sample-docs/IRS-form-1987.pdf")
p1, p2 = [page.elements for page in doc.pages]
```
`p1` and `p2` should contain a lot of elements with non-empty text.


MthwRobinson (approved):
LGTM!

PR #102, opened 2023-05-05
Added execution providers for TensorRT and CUDA

benjats07:
Adds execution providers for ONNX allowing use of GPU automatically.


qued (approved):
LGTM!

PR #103, opened 2023-05-05
Added ONNX model for detectron2

benjats07:
This PR adds ONNX model for detectron2. 
After some tests I conclude that the performance remains similar (~15% less sometimes) to the original detectron2.
Currently, we are keeping the support for original detectron2. In order to use the onnx model via API, the calls should be sent to http://localhost:8000/layout/detectron2_onnx/pdf or http://localhost:8000/layout/detectron2_onnx/image.

For example:
curl -X 'POST' 'http://localhost:8000/layout/detectron2_onn/pdf' \
-F 'file=@sample-docs/loremipsum-flat.pdf' \
-F force_ocr=true \\
 | jq -C | less -R
 
 Thanks to @qued for the modifications required to keep both models.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,7 @@
+## 0.4.5
+
+* Detectron2 dependency deleted
+
 ## 0.4.4
 
 * Fixed patches not being a package.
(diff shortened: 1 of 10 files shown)


qued (approved):
LGTM!

PR #104, opened 2023-05-07
Misc updates and tests

qued:
- Fixed interpretation of outputs, for example detectron bbox format is `(x1, y1, x2, y2)` instead of `(x1, y1, w, h)`
- Converted bbox coordinates to agree with original image dimensions
- Some (IMO) readability improvements
- Split `image_processing` function into pre and post processing, driven from the `predict` function
- Test updates (should all be passing right now)
- Type checking is passing

Have a look and see if you agree with the changes @benjats07 .


benjats07:
LGTM!
Thanks!

PR #105, opened 2023-05-08
feat: page annotation

qued:
Add capability for `PageLayout` to output annotated image.

Still to do:
- [x] Add test for `draw_bbox`
- [x] Add test for `annotate`
- [ ] Update or remove `draw_yolox_bounding_box` (not necessarily this PR)
- [ ] It would be nice to do something else with the `_COLORS` color map, maybe move the data to an external file, and maybe add comment labels to what color the RGB trio produces (not necessarily this PR)

#### Testing:
In a notebook do:
```python
doc = DocumentLayout.from_file("sample-docs/layout-parser-paper.pdf")
doc.pages[0].annotate()
```
The first page of the layout parser paper should appear with red bounding boxes around the elements.


unstructured_inference/inference/layout.py
```diff
@@ -192,6 +193,21 @@ def _get_image_array(self) -> Union[np.ndarray, None]:
             self.image_array = np.array(self.image)
         return self.image_array
 
+    def annotate(self, colors: Optional[Union[List[str], str]] = None) -> Image.Image:
+        """Annotates the elements on the page image."""
+        if colors is None:
+            colors = ["red" for _ in self.elements]
+        if isinstance(colors, str):
+            colors = [colors]
+        # If there aren't enough colors, just cycle through the colors a few times
```
benjats07, line 202: shouldn't go to the default color?

benjats07:
LGTM!

benjats07 (approved):

unstructured_inference/inference/layout.py
qued: We could do that, but if someone passed a colormap to get some differentiation between the bounding boxes, and then halfway through the page they switched to only red boxes, it would be a bit weird.

PR #106, opened 2023-05-09
fix: pip-compile test before dev

qued:
Since I made the requirements files tiered, with `test.txt` solved using the pins from `base.txt`, and `dev.txt` solved using the pins from `base.txt` and `test.txt`, it follows that the there needs to be an order to the `pip-compile`s: `base.txt -> test.txt -> dev.txt`. If they are done in a different order that means `dev.txt` might end up being solved relative to an updated pin list and an old pin list, which will likely be incompatible.

Since the `Makefile` generates the compiles in the order `base -> dev -> test`, this is exactly what happened. This fix corrects the order.

#### Testing:
Try `make pip-compile` on `main` branch, will likely fail unless the compiles are very recent.
Try `make pip-compile` on this branch. Should succeed.


benjats07 (approved):
LGTM

PR #107, opened 2023-05-12
enhancement: preserve image format

qued:
`PIL` records the filetype of the image when it's loaded, but the way we were handling the image destroyed the format metadata. This change keeps the image format metadata so that it can be recovered downstream.

#### Testing:
The following code should produce output `None` on the `main` branch and output `JPEG` on this branch.
```python
doc = DocumentLayout.from_image_file("sample-docs/loremipsum.jpg")
image = doc.pages[0].image
print(image.format)

```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,6 @@
-## 0.4.5-dev0
+## 0.4.5-dev1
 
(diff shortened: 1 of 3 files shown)


MthwRobinson (approved):
LGTM! This is great, will definitely want the file metadata downstream.

PR #112, opened 2023-05-18
fix: correct column ordering for multi-column documents

MthwRobinson:
### Summary

Addressed https://github.com/Unstructured-IO/unstructured/issues/356 . Updates the element ordering logic to account for multicolumn documents. The ordering logic is now as follows:

- Order the elements from top to bottom.
- From top to bottom:
  - If an element takes up the width of the page, add it to the elements list.
  - Otherwise, split the elements into columns based the `x` coordinates of the bounding box. Order the columns from left to right and elements within the columns from top to bottom. Add each element to the elements list in order.

### Testing

Run the following code, the order of the output should be correct when you compare it to the example file. Note, it seems as though the first paragraph is not detected, but that's an issue separate from this PR

```python

filename = "sample-docs/design-thinking.pdf"
doc = process_file_with_model(filename=filename, model_name=None)

for element in doc.pages[0].elements:
    print(element.text)
    print()
    input()
```


qued (approved):
LGTM! I definitely see the difference in the output order.

PR #118, opened 2023-05-19
fix: store page no

qued:
Page numbers weren't being correctly stored in the `PageLayout` objects. This is a fix so that page numbers are stored correctly.

```python
doc = DocumentLayout.from_file("layout-parser-paper.pdf")
print([page.number for page in doc.pages])

```
Output should be `[1, 2, 3, 4, ...]`.


benjats07 (approved):
LGTM

PR #119, opened 2023-05-19
fix: Exception handling for ONNX runtime on blank pages

benjats07:
This PR adds a try-catch during the inference with ONNX detectron2. 
A definitive solution would be to modify the ONNX model, however, doesn't seem like an easy change, and handling this error will allow the users to keep using the model.


qued (approved):
LGTM!

PR #120, opened 2023-05-30
fix: temporarily disable multicolumn logic

qued:
Temporarily short circuiting the new multi-column ordering logic after seeing examples of it misordering elements where it wasn't before.

#### Testing:
I added a test to verify that a problem layout is ordered correctly, so with line 56 in `ordering.py` commented out, the test `test_layout_order` in `test_layout` should fail, and otherwise it should succeed.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,8 +1,9 @@
-## 0.5.1-dev2
+## 0.5.1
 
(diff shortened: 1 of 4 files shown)


MthwRobinson (approved):
LGTM - that matches the original layout logic.

PR #125, opened 2023-06-06
build: add test for integration with unstructured to CI

qued:
Added to CI a test that grabs the latest version of `unstructured`, installs the ref version of `unstructured-inference` along with the other `unstructured` dependencies, and runs the ingest tests.

The idea is to catch errors or changes that cause issues downstream before we bump the inference version in `unstructured`.


MthwRobinson (approved):
LGTM!

PR #127, opened 2023-06-15
Fall back to pytesseract if paddleocr fails

rbiseck3:


cragwolfe (commented):
The changes make sense to me!

benjats07:
LGTM

benjats07 (approved):

unstructured_inference/__version__.py
benjats07, line 1: Probably we want 0.5.3-dev, but not sure if we are going to release this version.
What do you think @cragwolfe?

benjats07 (approved):

unstructured_inference/__version__.py
qued: Yeah this should either be 0.5.2 (if we are ready to release) or 0.5.2-dev2.

CHANGELOG.md
```diff
@@ -1,4 +1,4 @@
-## 0.5.3
+## 0.5.2-dev2
 
 * Configure fallback for OCR token if paddleocr doesn't work to use tesseract
```
qued: The pattern is to keep all changes consolidated until we do a release. So this section should be removed, the change should be listed in the `0.5.2-dev1` section, and then the `0.5.2-dev1` header should be bumped to `0.5.2-dev2`.

unstructured_inference/__version__.py
cragwolfe: i think we are ready to release! so that we can bring it into unstructured and then unstructured-api

PR #132, opened 2023-06-28
feat: large model

qued:
Added large model to accessible models. This required adding a new code path for this type of model. I compensated for the fact that elements from this model don't have location data by adding a new class `LocationlessLayoutElement`, but long term I think we should alter the `TextRegion` to have a `bbox` property instead of subclassing `Rectangle`.

#### Testing:
```python
import os

os.environ["UNSTRUCTURED_HF_TOKEN"] = "<enter hf token>"
model = get_model("large_model")
doc = DocumentLayout.from_file("sample-docs/loremipsum.pdf", element_extraction_model=model)
doc.pages[0].elements

```


ajjimeno (approved):
LGTM! Thanks Alan, it looks great. I checked the code and tested it on an AWS instance.

CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+## 0.5.2-dev0
```
cragwolfe, line 1: or is it release time? 😁

unstructured_inference/models/largemodel.py
```diff
@@ -0,0 +1,171 @@
...
+    DonutImageProcessor,
+    DonutProcessor,
+    VisionEncoderDecoderModel,
+)
+
+from unstructured_inference.inference.layoutelement import LocationlessLayoutElement
+from unstructured_inference.models.unstructuredmodel import UnstructuredElementExtractionModel
+
+MODEL_TYPES = {
+    "large_model": {
```
cragwolfe, line 18: can we call this something more specific, even `large_model_jun23` if not feeling especially inspired?
reason being, this will not be the last "large_model"

qued: I too don't think large_model is a good name.

unstructured_inference/models/largemodel.py
```diff
@@ -0,0 +1,171 @@
...
+    "Link",
+    "Misc",
+    "Field-Name",
+    "Value",
+    "Threading",
+    "Metadata",
+    "Form",
+]
+
+# NOTE(alan): Order matters with these
```
MthwRobinson, line 52: May be worth a comment on how the tags work, since it's not completely obvious (at least not to me)

unstructured_inference/models/largemodel.py
```diff
@@ -0,0 +1,171 @@
...
+            self.processor(
+                np.array(
+                    image,
+                    np.float32,
+                ),
+                return_tensors="pt",
+            ).pixel_values,
+            decoder_input_ids=torch.tensor([[0]]),
+            do_sample=True,
+            top_p=0.92,
```
MthwRobinson, line 105: Do any of these settings need to change by model? For instance between the large model and donut?

qued: Good question. @ajjimeno @LaverdeS @benjats07 ?

unstructured_inference/models/largemodel.py
qued: Expanded the note with more detail.

MthwRobinson (approved):
LGTM from my end as long as you're good to go on the settings question

unstructured_inference/models/largemodel.py
```diff
@@ -0,0 +1,171 @@
...
+]
+
+# NOTE(alan): Order matters with these
+SPECIAL_TAGS = ["<sep/>"] + [
+    f"<{open_or_close}s_{label_type}>" for label_type in LABEL_MAP for open_or_close in ("", "/")
+]
+
+
+class UnstructuredLargeModel(UnstructuredElementExtractionModel):
+    required_w: int = 1248
```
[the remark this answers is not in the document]
ajjimeno: The image gets rescaled. If the image is too large, it should be split. I think it should be fine with standard documents.

PR #135, opened 2023-06-29
fix: deterministic ordering

qued:
Updated ordering method to rely on upper-left x-coordinate in case of a y-coordinate tie.

### DIFF
--- a/unstructured_inference/inference/ordering.py
+++ b/unstructured_inference/inference/ordering.py
@@ -27,7 +27,7 @@ def order_layout(
     if len(layout) == 0:
         return []
 
-    layout.sort(key=lambda element: (element.y1, element.x1))
+    layout.sort(key=lambda element: (element.y1, element.x1, element.y2, element.x2))
     # NOTE(alan): Temporarily revert to orginal logic pending fixing the new logic
     # See code prior to this commit for new logic.
     return layout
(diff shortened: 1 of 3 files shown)


unstructured_inference/inference/ordering.py
```diff
@@ -27,7 +27,7 @@ def order_layout(
     if len(layout) == 0:
         return []
 
-    layout.sort(key=lambda element: element.y1)
+    layout.sort(key=lambda element: (element.y1, element.x1))
```
rbiseck3, line 30: Should this logic be moved to the unstructured repo anyways?

rbiseck3: Since we already have this logic in the unstructured repo: [unstructured/partition/pdf.py#L291](https://github.com/Unstructured-IO/unstructured/blob/main/unstructured/partition/pdf.py#L291)
```python
        sorted_page_elements = sorted(
            page_elements,
            key=lambda el: (
                el.coordinates[0][1] if el.coordinates else float("inf"),
                el.coordinates[0][0] if el.coordinates else float("inf"),
            ),
        )
```
it would benefit us to have a dedicated method to support sorting all page elements to maintain consistency across strategies.

cragwolfe (approved):
✅

PR #143, opened 2023-07-05
Table bug

ajjimeno:
Evaluated a fix for the table bug. This seems to appear with tables that are turned 90 degrees.


benjats07 (approved):
LGTM

PR #144, opened 2023-07-05
build: update connector installs

qued:
Update connector installs to keep current with `unstructured` library.


benjats07 (commented):
LGTM

benjats07 (approved):

PR #146, opened 2023-07-11
fix installation with python 3.8.17

benjats07:
This PR just makes pip-compile to requirements in -dev version.
With the current code is not possible to install unstructured along side unstructured-inference, `make install` on the last one return: 
`ERROR: Ignored the following versions that require a different python version: 1.25.0 Requires-Python >=3.9; 1.25.0rc1 Requires-Python >=3.9; 1.25.1 Requires-Python >=3.9
ERROR: Could not find a version that satisfies the requirement triton==2.0.0 (from versions: none)
ERROR: No matching distribution found for triton==2.0.0`

In order to check:
* git clone unstructured && cd unstructured
* make install
* git clone unstructured-inference && cd unstructured-inference
* make install  

Should install without issues.

### DIFF
--- a/requirements/base.txt
+++ b/requirements/base.txt
@@ -10,15 +10,13 @@ certifi==2023.5.7
     # via requests
 cffi==1.15.1
     # via cryptography
-charset-normalizer==3.1.0
+charset-normalizer==3.2.0
     # via
     #   pdfminer-six
     #   requests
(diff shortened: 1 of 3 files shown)


qued (approved):
LGTM, looks like this was dependabot's fault.

PR #149, opened 2023-07-12
Feat/detectron2 mask rcnn x101

benjats07:
This (Draft) PR adds a new detectron2 model under detectron2onnx class.

* Generate tighter bounding boxes for the elements.
* Works well with german
* Detects more elements than the previous model.
* Show less overlapping for regions.
* Model size goes from 166 Mb to 429 Mb (due to distinct architecture).
* In references pages behaves like the previous model
* Inference is faster than previous model 

For test:

`python test.py `

(This script will be removed before merging)

And will produce results in `results` folder.


qued:
Could you give examples where the new model is performing better? Like which document and page.

qued (approved):
LGTM, just one comment.

benjats07:
> Could you give examples where the new model is performing better? Like which document and page.

Sorry, my mistake...I just forget to add `test.py`, this script generates outputs ("result" folder) from some files in sample-docs and other files in unstructured repo.

cragwolfe:
Can we make this a proper release for ease of testing with unstructured? Should there be additional evaluation before making this the default model for layout detection?

benjats07:
After seeing the performance with the folder mentioned before don't think requires more testing, shows a similar performance to the previous model but with better bounding boxes. 
@pravin-unstructured also mentioned that have been testing the model and is faster than the previous one.

PR #151, opened 2023-07-14
enhancement: expose dpi

qued:
Exposed pdf-to-image conversion dpi parameter to higher level functions so this parameter can be set externally.

#### Testing:
```
doc = DocumentLayout.from_file("sample-docs/loremipsum.pdf", pdf_image_dpi=200)
print(doc.pages[0].image.height) # Should be 2200
doc = DocumentLayout.from_file("sample-docs/loremipsum.pdf", pdf_image_dpi=100)
print(doc.pages[0].image.height) # Should be 1100

```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
-## 0.5.6-dev0
+## 0.5.6-dev1
 
 * Warns users that Chipper is a beta model.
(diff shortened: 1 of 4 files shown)


MthwRobinson (approved):
LGTM!

PR #152, opened 2023-07-18
Fix/pdf2image too many open files

christinestraub:
### Summary
This PR is to fix the `too many open files` issue when loading `pdf` and to add functionality to set `page.image` to `None` to reduce memory usage.

* update `load_pdf` to return either Image objects or Image paths
* update `DocumentLayout.from_file` to open only one image
* add `image_metadata` attribute to PageLayout Class & set `page.image` to None to reduce memory usage
* add functionality to store pdf images for later use
* update the `annotate` and `_get_image_array` methods of `PageLayout` to get the image from the `image_path` property if the `image` property is `None`

### Testing
```
    filename = "sample-docs/pdf2image-memory-error-test-400p.pdf"
        filename,
        is_image=False,
        ocr_languages="eng",
        extract_tables=False,
        model_name=None,
    )

    print(layout.pages[0].elements)
```


benjats07 (approved):
LGTM, just needs the changelog fix :)

unstructured_inference/inference/layout.py
```diff
@@ -412,20 +450,26 @@ def load_pdf(
...
+        filename,
+        dpi=dpi,
+        output_folder=output_folder,
+        paths_only=path_only,
+    )
+
     return layouts, images
+
+
+def create_image_output_dir(
```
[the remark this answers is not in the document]
christinestraub: It's right

qued (commented):
Tested it just now -- on `main` I get the "too many open files" error when max file descriptors is set to 256, but on this branch I don't. However, running locally on this branch the process gets killed presumably due to running out of memory.

The code I ran was:
```python
doc = DocumentLayout.from_file("sample-docs/pdf2image-memory-error-test-400p.pdf")
```

Note the process is also killed running on `main`. I'll try to investigate this. The memory spike seems to be happening during the inference phase, and not during the PDF loading phase.

qued (approved):
Given that this solves the "too many open files" issue, and having tested that `annotate` still works, I'm going to approve this since the memory issue I mentioned also exists on `main`, making this a strict improvement over `main`. I'll just note that there still seems to be some sort of memory issue to resolve.

unstructured_inference/inference/layout.py
```diff
@@ -412,20 +450,26 @@ def load_pdf(
...
-        end_page = min(start_page + chunk_size - 1, total_pages)
-        with tempfile.TemporaryDirectory() as tmpdir:
-            chunk_images = pdf2image.convert_from_path(
-                filename,
-                dpi=dpi,
-                first_page=start_page,
-                last_page=end_page,
-                output_folder=tmpdir,
-            )
-            images += chunk_images
```
six5532one, line 430: Is this the part of the code that was suspected to cause "too many files open" errors? If the `with` context manager automatically closes the temp directory after each iteration, I don't understand how this would result in the error. Let me know if I'm missing something.

unstructured_inference/inference/layout.py
```diff
@@ -81,29 +82,43 @@ def from_file(
...
+                    number=i + 1,
+                    detection_model=detection_model,
+                    element_extraction_model=element_extraction_model,
+                    layout=layout,
+                    ocr_strategy=ocr_strategy,
+                    ocr_languages=ocr_languages,
+                    fixed_layout=fixed_layout,
+                    extract_tables=extract_tables,
+                )
+                pages.append(page)
```
six5532one, line 121: >running locally on this branch the process gets killed presumably due to running out of memory

Might `pages` be consuming too much memory here?

christinestraub:
> Tested it just now -- on `main` I get the "too many open files" error when max file descriptors is set to 256, but on this branch I don't. However, running locally on this branch the process gets killed presumably due to running out of memory.
> 
> The code I ran was:
> 
> ```python
> from unstructured_inference.inference.layout import DocumentLayout
> doc = DocumentLayout.from_file("sample-docs/pdf2image-memory-error-test-400p.pdf")
> ```
> 
> Note the process is also killed running on `main`. I'll try to investigate this. The memory spike seems to be happening during the inference phase, and not during the PDF loading phase.

@qued I've never seen the process killed due to running out of memory. The process took memory less than 1 GB (about 960 MB) at max when I tested.

unstructured_inference/inference/layout.py
christinestraub: @six5532one Yes, this part of the code causes the issue. The `with` context manager automatically closes the temp directory after each iteration, but the chunk image files added to the 'images' variable remain open even after deletion.

unstructured_inference/inference/layout.py
christinestraub: @six5532one @qued It should not be consuming too much memory here because I set the `image` property of the `PageLayout` instance to `None` to reduce the memory usage - https://github.com/Unstructured-IO/unstructured-inference/blob/fix/pdf2image-too-many-open-files/unstructured_inference/inference/layout.py#L325

unstructured_inference/inference/layout.py
six5532one: Thanks for explaining.

benjats07:
I tried the code and had no issues, the memory consumption was around 1 Gb and was tested on M2 chip. Notice that I tested on a brand new environment.

PR #153, opened 2023-07-19
enhancement: keep element type when merging

qued:
As outlined in [#927](https://github.com/Unstructured-IO/unstructured/issues/927) in the `unstructured` library, currently when we merge the extracted elements with the inferred elements, if it's determined that an element from the extracted set is the same element as an element in the inferred set, we use the extracted element because we trust that the text is correct. What would be better is to update the classification of the element using the inferred type, because this tends to be more accurate. That's what this PR is for.


MthwRobinson (approved):
LGTM!

qued:
Closing for now, after investigation it looks like the issue might be something else.

PR #154, opened 2023-07-21
build(deps): update detectron2

qued:
I noticed an issue that basic operations in this library were causing import errors. This turned out to be caused by the new version of `PIL` completing deprecation of certain references (specifically `PIL.Image.LINEAR`) used in the pinned version of `detectron2` that we've been using. Updating to the latest commit of `detectron2` fixes the problem.

#### Testing:
From a fresh environment, run `make install` from the `main` branch and observe that running `from unstructured_inference.inference.layout import DocumentLayout` causes an import error from within the `detectron2` code.

Subsequently, if you `pip uninstall detectron2`, switch to this branch, run `make install-detectron2`, and rerun the code, it should succeed.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,7 +1,8 @@
-## 0.5.6-dev1
+## 0.5.6-dev2
 
 * Warns users that Chipper is a beta model.
 * Exposed control over dpi when converting PDF to an image.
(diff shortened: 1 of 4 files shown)


MthwRobinson (approved):
LGTM!

PR #155, opened 2023-07-23
enhancement: cache named models

qued:
Fixes the memory issue that was discovered in #152.

* Caches named models as they are loaded to prevent reloading or creating multiple copies.
* Adds test that fails if model is initialized more than once (fails on `main`).

This should also address concerns that Chipper was being loaded multiple times @ajjimeno.

#### Testing:

The following code should run without crashing or the memory footprint exploding (takes a long time):
```python
DocumentLayout.from_file("sample-docs/pdf2image-memory-error-test-400p.pdf")
```

### DIFF
--- a/test_unstructured_inference/models/test_detectron2.py
+++ b/test_unstructured_inference/models/test_detectron2.py
@@ -17,14 +17,17 @@ class MockDetectron2LayoutModel:
 
 def test_load_default_model(monkeypatch):
     monkeypatch.setattr(detectron2, "Detectron2LayoutModel", MockDetectron2LayoutModel)
+    monkeypatch.setattr(models, "models", {})
 
     with patch.object(detectron2, "is_detectron2_available", return_value=True):
         model = models.get_model("detectron2_lp")
 
+    print(type(model))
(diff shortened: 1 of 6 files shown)


test_unstructured_inference/models/test_detectron2.py
```diff
@@ -17,14 +17,17 @@ def detect(self, x):
 
 def test_load_default_model(monkeypatch):
     monkeypatch.setattr(detectron2, "Detectron2LayoutModel", MockDetectron2LayoutModel)
+    monkeypatch.setattr(models, "models", {})
 
     with patch.object(detectron2, "is_detectron2_available", return_value=True):
         model = models.get_model("detectron2_lp")
 
+    print(type(model))
```
MthwRobinson, line 25: Reminder to remove this print statement

MthwRobinson (approved):
This setup looks good to me and is similar to what we had a while back when we just had the one model

MthwRobinson:
cc @benjats07 since I think you brought up multiple model loads before too

PR #158, opened 2023-07-26
Update pip in makefile

shreyanid:
Update pip version to latest (23.2.1)

Test: `make install` has no pip warnings


cragwolfe:
should probably do the "make pip-compile" dance here as well, and bump to `0.5.8-dev0`

cragwolfe (approved):
LGTM!

PR #160, opened 2023-07-28
docs: Update README.md- In 'Using a non-default model', argument name updated to 'detection_model'

anukurian16:
In the section 'Using a non-default model', the argument name passed to DocumentLayout.from_file() must be 'detection_model' instead of 'model'


cragwolfe (approved):

cragwolfe:
thanks, @anukurian16 !

PR #165, opened 2023-08-04
fix: handle errors from Tesseract

awalker4:
Certain regions of a document are failing ocr with this error:

`pytesseract.pytesseract.TesseractError: (-8, 'Estimating resolution as 1250')` (for some value)

When I try the same region on the CLI, I get:

```
$ tesseract bad_tile.jpeg output
Estimating resolution as 1813
Floating point exception
```

Whatever the root cause, let's catch this error and return an empty string.

### DIFF
--- a/unstructured_inference/inference/elements.py
+++ b/unstructured_inference/inference/elements.py
@@ -269,7 +269,11 @@ def ocr(text_block: TextRegion, image: Image.Image, languages: str = "eng") -> s
     agent = tesseract.ocr_agents.get(languages)
     if agent is None:
         raise RuntimeError("OCR agent is not loaded for {languages}.")
-    return agent.detect(cropped_image)
+
+    try:
+        return agent.detect(cropped_image)
+    except tesseract.TesseractError:
+        return ""
 
 
 def needs_ocr(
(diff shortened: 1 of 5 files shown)


awalker4:
Looks like those ingest tests have been failing on main for a while, so I'm assuming we can ignore here.

cragwolfe:
> Looks like those ingest tests have been failing on main for a while, so I'm assuming we can ignore here.

Yeah, that just indicates the current version of unstructured-inference would produce different outputs if the ingests tests are run. Hopefully, they are good different outputs!

shreyanid (approved):
Code looks good to me! Tested in docker to reproduce the issue, and this change results in a 200 status for otherwise failing documents.

unstructured_inference/inference/elements.py
```diff
@@ -269,7 +269,11 @@ def ocr(text_block: TextRegion, image: Image.Image, languages: str = "eng") -> s
     agent = tesseract.ocr_agents.get(languages)
     if agent is None:
         raise RuntimeError("OCR agent is not loaded for {languages}.")
-    return agent.detect(cropped_image)
+
+    try:
+        return agent.detect(cropped_image)
+    except tesseract.TesseractError:
```
cragwolfe, line 275: add a `logger.warn("<something", exc_info=True)`

PR #167, opened 2023-08-08
Added logic for ordering two-column documents

benjats07:
This PR adds `order_two_column_document` and `order_two_column_page` for ordering elements in reading order.

In order to test you can execute:
```
from unstructured_inference.inference import elements, layout
file = "sample-docs/Dense-Passage-Retrieval-for-Open-Domain-Question-Answering.pdf"
detectron2_onnx = get_model("detectron2_onnx")
doc = layout.DocumentLayout.from_file(file, detection_model=detectron2_onnx)
ordered_document = order_two_column_document(doc)
pages_element = [p.elements for p in ordered_document.pages]
elements = sum(pages_element, [])
content = " ".join([e.text for e in elements])
```
and check that `content` now have the ordering is very similar to the actual document.


unstructured_inference/inference/ordering.py
```diff
@@ -31,3 +33,43 @@ def order_layout(
...
+    new_bounding_boxes_ix.extend(left_boxes)
+    new_bounding_boxes_ix.extend(right_boxes)
+    return new_bounding_boxes_ix
+
+
+def order_two_column_document(document: DocumentLayout) -> DocumentLayout:
+    """Orders all pages in the document assuming the existence of two columns"""
+    for page in document.pages:
+        ordered_elements = order_two_column_page(
+            page.elements,
```
qued, line 71:
```suggestion
    for page in document.pages:
        bbox_elements = [el for el in page.elements if isinstance(el, LayoutElement)]
        no_bbox_elements = [el for el in page.elements if not isinstance(el, LayoutElement)]
        ordered_elements = order_two_column_page(
            cast(List[LayoutElement], bbox_elements),
```

unstructured_inference/inference/ordering.py
```diff
@@ -31,3 +33,43 @@ def order_layout(
...
+            width_page=page.image_metadata["width"],
+        )
+        page.elements = ordered_elements
```
qued, line 74:
```suggestion
        page.elements = ordered_elements + no_bbox_elements
```

unstructured_inference/inference/ordering.py
```diff
@@ -31,3 +34,51 @@ def order_layout(
...
+    new_bounding_boxes_ix = []
+    for i, bbox in enumerate(elements):
+        x_min = bbox.x1
+        x_max = bbox.x2
+        if x_min < vertical_line_x and x_max < vertical_line_x:
+            left_boxes.append(bbox)
+        elif x_min > vertical_line_x and x_max > vertical_line_x:
+            right_boxes.append(bbox)
+        else:
+            both_sided_boxes.append(bbox)
```
qued, line 58: I'm worried things like page numbers, headers, and footers would get classified as left boxes and right boxes and get put in a weird order.

qued: We might need to identify margins to avoid that. Not that that's easy.

qued (commented):
A few comments.

unstructured_inference/inference/ordering.py
```diff
@@ -31,3 +34,51 @@ def order_layout(
...
+    # Determine the order of the bounding boxes
+    left_boxes = []
+    right_boxes = []
+    both_sided_boxes = []
+
+    new_bounding_boxes_ix = []
+    for i, bbox in enumerate(elements):
+        x_min = bbox.x1
+        x_max = bbox.x2
+        if x_min < vertical_line_x and x_max < vertical_line_x:
```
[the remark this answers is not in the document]
benjats07: Oh, I see your point. 
Yes, with x_max<vertical_line_x could be enough.
I'm wondering if this relation between x1 and x2 is enforced in code.

unstructured_inference/inference/ordering.py
benjats07: Yeah, that elements will be categorized as any of those, but ... where should be placed? They aren't in a column

qued: Off the top of my head, I'm thinking headers first, then body text by column, with left and right margin stuff included as body text (this isn't quite right for the arxiv watermarks in the left margin, which is more like header text, but I'm thinking of cases where margin text is used for notes on the text it's adjacent to, and I think including it inline would work), then footers last. Page numbers included in the header/footer is ok I think.

cragwolfe:
Closing since stale but this may be re-opened, see https://github.com/Unstructured-IO/unstructured/pull/1041#issuecomment-1711066334

PR #168, opened 2023-08-09
build: revert default model, cut release

cragwolfe:
For reasons not yet entirely clear, using the model introduced in:
https://github.com/Unstructured-IO/unstructured-inference/commit/04b2e6fe5bf9506551714f11ffdbd1abfc28ea13
is causing regressions with how elements are labeled (esp. a lot of `UncategorizedText` vs. `NarrativeText`).

Until that is resolved, keep the old model as the default base model (but this may be overridden, as usual).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,6 @@
-## 0.5.8-dev2
+## 0.5.8
 
-* Add alternative architecture for detectron2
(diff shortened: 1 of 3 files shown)


benjats07 (approved):
LGTM

qued (approved):
LGTM!

PR #171, opened 2023-08-15
enhancement: warn if low dpi chipper

qued:
Chipper is best used when images are rendered at 300 dpi or above. This PR adds a warning if Chipper is used with DPI less than 300.

#### Testing:

Run the following code:
```python
process_file_with_model(filename="sample-docs/loremipsum.pdf", model_name="chipper", pdf_image_dpi=299)

```
You should get 2 warnings, the first being that Chipper should be used with DPI >= 300.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,7 @@
+## 0.5.11-dev0
+
+* Add warning when chipper is used with < 300 DPI
+
 ## 0.5.10
 
 * Implement full-page OCR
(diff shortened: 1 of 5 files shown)


ajjimeno:
It looks good to me. I run the test locally and it worked. Not sure what might be going on in CI.

ajjimeno (approved):
Looks good to me. It worked locally, not sure why it fails in PR.

PR #172, opened 2023-08-15
Fix for wrong type assignation at ingest-test

benjats07:
This PR changes some thresholds for getting the ingest test correct at `unstructured`.

### DIFF
--- a/unstructured_inference/inference/layout.py
+++ b/unstructured_inference/inference/layout.py
@@ -28,6 +28,9 @@ from unstructured_inference.inference.layoutelement import (
 from unstructured_inference.inference.ordering import order_layout
 from unstructured_inference.logger import logger
 from unstructured_inference.models.base import get_model
+from unstructured_inference.models.detectron2onnx import (
+    UnstructuredDetectronONNXModel,
+)
 from unstructured_inference.models.unstructuredmodel import (
     UnstructuredElementExtractionModel,
     UnstructuredObjectDetectionModel,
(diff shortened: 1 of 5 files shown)


qued (commented):
I'm worried we're dropping elements with the new thresholds. For example in the ingest tests document, `2023-Jan-economic-outlook.pdf`, page 6: 

Old thresholds: 
image

New thresholds:
image

Page 9 is another example of this.

This is after merging `main` into this branch and comparing this branch to `main`.

unstructured_inference/inference/layout.py
```diff
@@ -262,15 +265,33 @@ def get_elements_with_detection_model(
             raise ValueError("Invalid OCR mode")
 
         if self.layout is not None:
+            kwargs = {}
```
cragwolfe, line 268:
```suggestion
            threshold_kwargs = {}
```
a bit of a nit, but let's make it clear what this kwargs variable is for rather than the kitchen sink

cragwolfe:
I'm not sure the ingest tests diff's are an improvement so far. But, the default model has not changed, right? The existing/prior default model `detectron2_onnx` ideally would not have any changes in output unless those changes are clearly better.

benjats07:
> I'm not sure the ingest tests diff's are an improvement so far. But, the default model has not changed, right? The existing/prior default model `detectron2_onnx` ideally would not have any changes in output unless those changes are clearly better.

Yeah, the default model keep being `detectron2_onnx`. I think the changes are due to the implementation of `merge_inferred_layout_with_extracted_layout` and `merge_inferred_layout_with_ocr_layout` since the model remains the same.

cragwolfe (approved):
Default `detectron2_onnx` behaviour is unchanged, per CI test_ingest tests.

In theory, the thresholds set in this PR should be better for `detectron2_mask_rcnn`(not the default model), but this is not proven as part of this PR.

Approved since clearly no issues with the default path.

PR #173, opened 2023-08-15
Klaijan/add tiff image file support case in from_image_file function

Klaijan:
Modify `from_image_file` from class `DocumentLayout` to support reading multipage TIFF. 

Note:
The function extracts 2 pages TIFF at 1.24 ms, while extracting 2 jpgs and pngs at 1.54 and 1.32 ms respectively.

### DIFF
--- a/unstructured_inference/inference/layout.py
+++ b/unstructured_inference/inference/layout.py
@@ -10,7 +10,7 @@ import pdf2image
 import pytesseract
 from pdfminer import psparser
 from pdfminer.high_level import extract_pages
-from PIL import Image
+from PIL import Image, ImageSequence
 from pytesseract import Output
 
 from unstructured_inference.inference.elements import (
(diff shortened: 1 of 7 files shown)


cragwolfe (commented):
this will need to be a non-dev release to be include in `unstructured`, so please create a new CHANGELOG section for 0.5.12 and add your one bullet to it. likewise, update `unstructured_inference/__version__.py` to match.

unstructured_inference/inference/layout.py
```diff
@@ -143,26 +143,46 @@ def from_image_file(
         try:
             image = Image.open(filename)
             format = image.format
-            image = image.convert("RGB")
+            if format == "TIFF":
+                images = [im.convert("RGB") for i, im in enumerate(ImageSequence.Iterator(image))]
+            else:
+                image = image.convert("RGB")
```
qued, line 149: I think instead of handling `TIFF` as a separate case below, you could populate `images` with a singleton here:
```suggestion
                images = [image.convert("RGB")]
```
... and then below iterate over `images`.

qued (commented):
Overall looks great! I just have one requested change related to the code structure and a request for a test.

qued:
We should also have a test for `TIFF`. Probably we could just have an example `TIFF` and add on to an existing image test.

unstructured_inference/inference/layout.py
Klaijan: Fixed!

qued (approved):
LGTM!

cragwolfe (changes requested):
this will need to be a non-dev release to be include in unstructured, so please create a new CHANGELOG section for 0.5.13 and add your one bullet to it. likewise, update `unstructured_inference/__version__.py` to match.

cragwolfe (approved):

PR #174, opened 2023-08-16
fix: make DPI optional with default None

qued:
Made `pdf_image_dpi` default `None` to allow for more complex handling upstream.

### DIFF
--- a/unstructured_inference/inference/layout.py
+++ b/unstructured_inference/inference/layout.py
@@ -367,7 +367,7 @@ def process_data_with_model(
     ocr_mode: str = "entire_page",
     fixed_layouts: Optional[List[Optional[List[TextRegion]]]] = None,
     extract_tables: bool = False,
-    pdf_image_dpi: int = 200,
+    pdf_image_dpi: Optional[int] = None,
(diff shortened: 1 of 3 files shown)


unstructured_inference/inference/layout.py
```diff
@@ -397,11 +397,13 @@ def process_file_with_model(
...
     fixed_layouts: Optional[List[Optional[List[TextRegion]]]] = None,
     extract_tables: bool = False,
-    pdf_image_dpi: int = 200,
+    pdf_image_dpi: Optional[int] = None,
 ) -> DocumentLayout:
     """Processes pdf file with name filename into a DocumentLayout by using a model identified by
     model_name."""
 
+    if pdf_image_dpi is None:
+        pdf_image_dpi = 200
```
cragwolfe, line 406: should this get set to 300 if 

    model_name == "chipper"
    
?

qued: Incoming PR that does that upstream in `unstructured`.

cragwolfe (approved):
LGTM!

PR #178, opened 2023-08-16
fix: Fix pdfminer error when using process_data_with_model

awalker4:
When a pdf page doesn't have much data, it may get buffered in the write to a tempfile. If this happens, we'll hit an error reading the file back. Open to suggestions for a way to unit test this - I was creating some test files with pypdf but I couldn't trigger the error.

Steps to verify:
[bad_page.pdf](https://github.com/Unstructured-IO/unstructured-inference/files/12363180/bad_page.pdf)

* With the attached file, run this snippet on main and confirm that there's an exception - `PDFSyntaxError: No /Root object! - Is this really a PDF?`
* Run it on this branch and confirm the error goes away

```
from unstructured_inference.inference.layout import process_data_with_model

filename = "bad_page.pdf"

with open(filename, 'rb') as f:
    process_data_with_model(f, model_name="detectron2_lp")
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,7 @@
+## 0.5.12
+
+* Fix a pdfminer error when using `process_data_with_model`
+
 ## 0.5.11
 
 * Add warning when chipper is used with < 300 DPI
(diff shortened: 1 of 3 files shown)


qued (approved):
LGTM!

PR #182, opened 2023-08-21
Feat: onnx quantization

benjats07:
This PR adds "yolox_quantized" model as an alternative detection model, leveraging a similar speed to detectron2_onnx.
Also adds "detectron2_quantized" model as a quantized model for "detectron2_onnx".
Both models are generated at first usage.


CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+## 0.5.20-dev0
```
cragwolfe, line 1: let's make this a release (no `-dev0` prefix)

benjats07: Done!

cragwolfe:
Note: re-running test_ingests to get fresh diff against unstructured test-ingest fixtures:
https://github.com/Unstructured-IO/unstructured-inference/actions/runs/6045358745/job/16405565118?pr=182

cragwolfe:
I dug into the differences in ingest outputs here:
https://docs.google.com/document/d/1Ur0YABBvWLDXEVLJl22kO_0hgGCOgXfnu-eDEuJ3Q2c/edit?usp=sharing

While it seems table extraction is better, the overall results seem a bit worse, qualitatively.

unstructured_inference/models/base.py
```diff
@@ -23,7 +23,7 @@
     UnstructuredYoloXModel,
 )
 
-DEFAULT_MODEL = "detectron2_onnx"
+DEFAULT_MODEL = "yolox_quantized"
```
badGarnet, line 26: maybe we should use a config pattern that allows this to be set via env variable?

benjats07:
> I dug into the differences in ingest outputs here: https://docs.google.com/document/d/1Ur0YABBvWLDXEVLJl22kO_0hgGCOgXfnu-eDEuJ3Q2c/edit?usp=sharing
> 
> While it seems table extraction is better, the overall results seem a bit worse, qualitatively.

The ordering of the shouldn't be affected by the model change. Wondering if the algorithm changed at some point.

cragwolfe:
> The ordering of the shouldn't be affected by the model change. Wondering if the algorithm changed at some point.

"shouldn't" perhaps but it clearly is. This is against the unstructured library is up to date with `main` on unstructured-inference -- it is not the sorting algo, but the inputs to the sorting algo that is changing the results here.

cragwolfe:
Perhaps we should update this as not the default model until other quality issues are sorted out, and consumers can still specify `model_name=yolox_quantized` if they are focused on Table identification?

benjats07:
> > The ordering of the shouldn't be affected by the model change. Wondering if the algorithm changed at some point.
> 
> "shouldn't" perhaps but it clearly is. This is against the unstructured library is up to date with `main` on unstructured-inference -- it is not the sorting algo, but the inputs to the sorting algo that is changing the results here.

I'm running some tests, and seems like the integration of OCR elements disorders some things (I found some texts being added in cross-order), so isn't the model. On the other way, with that algorithm, a lot of text boxes are recovered (Things that YoloX is missing)

benjats07:
> Perhaps we should update this as not the default model until other quality issues are sorted out, and consumers can still specify `model_name=yolox_quantized` if they are focused on Table identification?

Agree, because solving all the problems at the same time will be very time-consuming.

unstructured_inference/models/base.py
benjats07: @cragwolfe, what do you think? 
I can get this string from an environment variable, if isn't present I can keep the model defined in this file.

cragwolfe: a change of config pattern is better a separate PR in the first place, don't need to mix that into this PR

.github/workflows/ci.yml
```diff
@@ -107,6 +107,7 @@ jobs:
       uses: actions/checkout@v3
       with:
         repository: 'Unstructured-IO/unstructured'
+        ref: 'feat/enhance-yolox-types-integration'
```
cragwolfe, line 110: remove this line?

unstructured_inference/models/yolox.py
```diff
@@ -66,6 +87,7 @@ def initialize(self, model_path: str, label_map: dict):
                 "CPUExecutionProvider",
             ],
         )
+
         self.layout_classes = label_map
 
     def image_processing(
```
cragwolfe, line 93: btw, given the doc string this function should likely be called `yolox_` something (image_to_layout?).

benjats07: Probably. We don't have a convention for that kind of methods. Detectron2 use `predict`, wouldn't harm to change the naming in another PR

cragwolfe (approved):

PR #183, opened 2023-08-22
fix: handle potential errors from tesseract call

### LINKED ISSUE #179: bug/another instance of TesseractError - Estimating resolution as x
See screenshot. We need to catch errors from Tesseract in the new `entire_page` ocr path. This [line](https://github.com/Unstructured-IO/unstructured-inference/blob/de19ace101632ed5a8b8e0fed62cef0795fd55ac/unstructured_inference/inference/layout.py#L243).

Screenshot 2023-08-18 at 11 49 18 AM

awalker4:
We've seen a 500 error in `unstructured-api` due to an uncaught TesseractError in the `entire_page` path. I can't reproduce it, but we can at least add a try catch. The last fix was too aggessive, which we're tracking [here][https://github.com/Unstructured-IO/unstructured/issues/1086], so we may need to adjust this fix as well.

Closes #179

### DIFF
--- a/unstructured_inference/inference/layout.py
+++ b/unstructured_inference/inference/layout.py
@@ -246,12 +246,16 @@ class PageLayout:
         if self.ocr_mode == "individual_blocks":
             ocr_layout = None
         elif self.ocr_mode == "entire_page":
-            ocr_data = pytesseract.image_to_data(
-                self.image,
-                lang=self.ocr_languages,
-                output_type=Output.DICT,
-            )
-            ocr_layout = parse_ocr_data(ocr_data)
+            ocr_layout = None
+            try:
+                ocr_data = pytesseract.image_to_data(
+                    self.image,
+                    lang=self.ocr_languages,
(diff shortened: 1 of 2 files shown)


unstructured_inference/inference/layout.py
```diff
@@ -246,12 +246,16 @@ def get_elements_with_detection_model(
...
+            ocr_layout = None
+            try:
+                ocr_data = pytesseract.image_to_data(
+                    self.image,
+                    lang=self.ocr_languages,
+                    output_type=Output.DICT,
+                )
+                ocr_layout = parse_ocr_data(ocr_data)
+            except pytesseract.pytesseract.TesseractError:
+                logger.warning("TesseractError: Skipping page")
```
cragwolfe, line 258: add `, exc_info=True` so there is a chance at debugging these issues?
actually same for the other try/catch in https://github.com/Unstructured-IO/unstructured-inference/pull/165/files

cragwolfe (approved):

PR #184, opened 2023-08-22
fix: use temporary instead of fixed directories for storing images of pdfs being processed

newelh:
**Summary**

- Replaces using a created directory for storing image outputs with a temporary directory
- Deprecates `create_image_output_dir` method
- Adds hot-loading for annotating images because images are no longer stored long-term in a directory
- Adds a document_filename keyword arg to the PageLayout to enable hot-loading

**Tests**

Removes tests associated with `create_image_output_dir`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,8 @@
+## 0.5.15-dev0
+
+* Fix to no longer create a directory for storing processed images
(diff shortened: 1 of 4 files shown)


cragwolfe (approved):
👏 👏

PR #185, opened 2023-08-25
fix: Set OMP_THREAD_LIMIT for better tesseract performance

awalker4:
I've spent some time playing with this var, and I came up with [this gist](https://gist.github.com/awalker4/8581d76d373c1bc51e0f2676a6ad816c). I ran this on a 4 core EC2 instance. Processing 3 pages without the limit takes 153s. With the limit is 5s 😍 . When the number of pages is higher than number of cores, it just hangs without this var.

### DIFF
--- a/unstructured_inference/models/tesseract.py
+++ b/unstructured_inference/models/tesseract.py
@@ -1,3 +1,4 @@
+import os
 from typing import Dict
 
 import pytesseract
@@ -8,6 +9,10 @@ from unstructured_inference.logger import logger
 ocr_agents: Dict[str, TesseractAgent] = {}
 
 TesseractError = pytesseract.pytesseract.TesseractError
+
+# Force tesseract to be single threaded,
+# otherwise we see major performance problems
+os.environ["OMP_THREAD_LIMIT"] = "1"
 
 
 def load_agent(languages: str = "eng"):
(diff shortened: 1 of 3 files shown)


tabossert:
Good find!

unstructured_inference/models/tesseract.py
```diff
@@ -9,6 +10,10 @@
 
 TesseractError = pytesseract.pytesseract.TesseractError
 
+# Force tesseract to be single threaded,
+# otherwise we see major performance problems
+os.environ["OMP_THREAD_LIMIT"] = "1"
```
cragwolfe, line 15: we should only set this if not already defined.
great find, though

cragwolfe (approved):

awalker4:
Just one more data point - instead of calling `ocr()` directly, I tried three threads doing `proccess_data_with_model` on layout-parser-paper.pdf. With the env var set, all 3 finish in ~100s. Without, we're around 500s.

```
Took 100.30185661300493 seconds
Took 102.10554121700989 seconds
Took 104.54798686499998 seconds
Total: 104.54726779799967 seconds

In [7]: del os.environ["OMP_THREAD_LIMIT"]

In [8]: run_ocr_threads(3)
Took 491.19365794499754 seconds
Took 502.786950402995 seconds
Took 502.9147502150008 seconds
Total: 502.91452509099327 seconds
```

Issue #176, opened 2023-08-16
enhancement: Elements detected by OCR should supplement detected layout


qued:
Similar to how elements extracted from a PDF are used to supplement the detected layout, elements from the full page OCR that do not have a corresponding element in the detected layout should be added to the output elements.

The merge process between extracted and detected elements can be used as a rough guide.

There is an expectation that at least the text of the `Image` elements lost in https://github.com/Unstructured-IO/unstructured-inference/pull/175 (and can be seen in the ingest-tests fixtures update [here](https://github.com/Unstructured-IO/unstructured/pull/1142/files) ) will be "revived" by the PR that addresses this issue.

If that's not the case, we'll need to revisit the current embedded-Image skipping logic (e.g. perhaps only skip "very large" embedded Images), and this should be tracked in a follow-on PR.

PR #186, opened 2023-08-29
Feat/176 elements detected by ocr supplement detected layout

christinestraub:
Closes [#176](https://github.com/Unstructured-IO/unstructured-inference/issues/176).
### Summary
- Add functionality to annotate any layout(extracted, inferred, OCR) on a page
- Add a Python script and a Jupyter Notebook for annotating layout elements
- Add functionality to get layout elements from OCR regions
- Add functionality to supplement detected layout with elements from the full-page OCR
- Add the utility function `annotate_layout_elements()`
- Add a Python script and a Jupyter Notebook to evaluate the feature implemented in this branch
### Testing
```

# pdf
doc = DocumentLayout.from_file("sample-docs/patent-1p.pdf")

# image
doc = DocumentLayout.from_image_file("sample-docs/layout-parser-paper-fast.jpg")
```
### Evaluation
A Python script and a Jupyter Notebook were added for evaluation.
```
PYTHONPATH=. python examples/ocr_layout_supplement/ocr_layout_supplement.py <file_path> <file_type>
```
For example,
```
PYTHONPATH=. python examples/ocr_layout_supplement/ocr_layout_supplement.py sample-docs/patent-1p.pdf pdf

### CODE REVIEW

**Comment by cragwolfe** [comment]
CC @badGarnet

**Comment by cragwolfe** [comment]
OK, the results for the IRS doc in the test_ingest fixture, which you can see the original [here](https://utic-dev-tech-fixtures.s3.us-east-2.amazonaws.com/ingest-test-doc-copies/azure/IRS-form-1987.pdf) ... are leaps and bounds better than what we had before 🎉 ! :
image
so much missing text recovered. And XYcut seems to be mostly ordering things pretty well in natural reading order.

---

Speaking of xycut, i'm a bit nervous that these "hyper granular" bounding boxes coming from OCR rather than bboxes layout detection actually would get ordered better with the existing xycut as is, rather than doing what https://github.com/Unstructured-IO/unstructured/issues/1209 suggests. In any case, that's a very minor point for the time being.

image

**Review by cragwolfe — APPROVED** [review]

File: unstructured_inference/utils.py [hunk 1]
```diff
@@ -39,3 +53,75 @@ def __iter__(self) -> Iterator:
...
+
+    for idx, page in enumerate(doc.pages):
+        for action_type, action_value in annotation_data_map.items():
+            img = page.annotate(annotation_data=action_value)
+            output_f_path = os.path.join(
+                output_dir_path,
+                f"{output_f_basename}_{idx + 1}_{action_type}.jpg",
+            )
+            if result == ANNOTATION_RESULT_WITH_IMAGE:
+                write_image(img, output_f_path)
```
cragwolfe, line 125:
```suggestion
                write_image(img, output_f_path)
                print(f"wrote {output_f_path}")
```
makes it easier to find the outputs when running examples/ocr_layout_supplement/ocr_layout_supplement.py

unstructured_inference/visualize.py
```diff
@@ -1,21 +1,23 @@
 # Copyright (c) Megvii Inc. All rights reserved.
 # Unstructured modified the original source code found at
 # https://github.com/Megvii-BaseDetection/YOLOX/blob/ac379df3c97d1835ebd319afad0c031c36d03f36/yolox/utils/visualize.py
+from typing import Optional, Union
 
 import cv2
+import matplotlib.pyplot as plt
```
qued, line 7: `matplotlib` is already in our dependencies via `pycocotools`, but to be safe we should add it explicitly to `requirements/base.in` if we're going to import it directly. After adding it, run `make pip-compile` to recompile dependencies.

qued (changes requested):
Over LGTM, a few minor things, and then we can merge. This will be a huge upgrade!

unstructured_inference/utils.py
christinestraub: Sounds good!

unstructured_inference/constants.py
```diff
@@ -0,0 +1,6 @@
+OCR_MODE_INDIVIDUAL_BLOCKS = "individual_blocks"
+OCR_MODE_FULL_PAGE = "entire_page"
```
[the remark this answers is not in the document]
christinestraub: Sounds good!

unstructured_inference/constants.py
```diff
@@ -0,0 +1,6 @@
...
+SUBREGION_THRESHOLD_FOR_OCR = 0.5
+
+ANNOTATION_RESULT_WITH_IMAGE = "image"
+ANNOTATION_RESULT_WITH_PLOT = "plot"
```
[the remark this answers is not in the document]
christinestraub: Sounds good!

unstructured_inference/visualize.py
christinestraub: Sounds good!

cragwolfe (changes requested):
hold up on the matplotlib for dependencies.
this is really like a test.in requirement, right? i.e., it is never going to be used in the context of `partition` from unstructured.

christinestraub:
> hold up on the matplotlib for dependencies. this is really like a test.in requirement, right? i.e., it is never going to be used in the context of `partition` from unstructured.

Yes, I also think it is never going to be used in the context of `partition` from unstructured. For now, it is only used for visualization for testing purposes.
Do we need to add it to `requirements/test.in` then? @cragwolfe

christinestraub:
The `matplotlib` is not going to be used in the context of `partition` from unstructured but it would be imported into layout.py via [`from unstructured_inference.visualize import draw_bbox`](https://github.com/Unstructured-IO/unstructured-inference/blob/main/unstructured_inference/inference/layout.py#L39). IMO, it would be good to add `matplotlib` explicitly to requirements/base.in to be safe, as @qued mentioned.

cragwolfe (approved):

qued (approved):
LGTM!

PR #196, opened 2023-09-05
Feat/enhanced detected types detectron

benjats07:
This PR makes a slight adjustment to confidence thresholds in the detectron2_onnx model in order to detect more element types.

The enhancements could be checked at ingestion_tests.

### DIFF
--- a/unstructured_inference/models/detectron2onnx.py
+++ b/unstructured_inference/models/detectron2onnx.py
@@ -37,7 +37,7 @@ MODEL_TYPES: Dict[Optional[str], Union[LazyDict, dict]] = {
             "model.onnx",
         ),
         label_map=DEFAULT_LABEL_MAP,
-        confidence_threshold=0.8,
+        confidence_threshold=0.5,
     ),
     "detectron2_quantized": {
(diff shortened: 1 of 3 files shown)


cragwolfe:
if the ingest tests results are better, this should be an actual release so we can get it into prod.

tabossert (approved):

mallorih (approved):

unstructured_inference/models/detectron2onnx.py
```diff
@@ -37,7 +37,7 @@
             "model.onnx",
         ),
         label_map=DEFAULT_LABEL_MAP,
-        confidence_threshold=0.8,
+        confidence_threshold=0.5,
```
badGarnet, line 40: can we make this configurable via a plain text file/env variable? It feel like we may want to have that ability ourselves at least. And how knows maybe in deployment we want to tweak it for different customers.

benjats07: Yeah, I was thinking about the same, but probably is better to add a new issue to record that is a need

PR #197, opened 2023-09-06
Fix/wrong ordering on supplementary ocr

benjats07:


cragwolfe:
i believe from the point of view of the `unstructured` library, this "intermediate" ordering won't matter much since a significant reordering happens as part of https://github.com/Unstructured-IO/unstructured/pull/1161/files#diff-cefa2d296ae7ffcf5c28b5734d5c7d506fbdb225c05a0bc27c6b755d5424ffda

benjats07:
Yeah, however, is changing the correct order of some tests, the one I've been tracking for checking is the `test_unstructured_ingest/download/s3/small-pdf-set/Silent-Giant-(1).pdf` file.  The expected output shows:

"text": "Nuclear energy is a proven solution ...
"text": "Energy is the essential agent ...
"text": "Nuclear energy has shown ...

and should be:

"text": "Energy is the essential agent ...
"text": "Nuclear energy is a proven solution ...
"text": "Nuclear energy has shown ...

But, I'm noting that the result of the ingest-test for this file is distinct from than `partition` call with the same file. (Partition call shows the content correctly)

PR #199, opened 2023-09-06
sort tokens before use

badGarnet:
This PR changes the logic inside table detection so use sorted tokens as input into `recognize` function. The tokens are sorted top down then left to right.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,7 @@
+## 0.5.23
+
+* use sorted tokens (top down then left to right) in table layout recognition
+
 ## 0.5.22
 
 * Add object-detection classification probabilities to LayoutElement for all currently implemented object detection models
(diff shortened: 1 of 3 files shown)


ajjimeno:
I have been checking the test errors and I have the impression that the sorting is somehow producing worst outcomes for the tables in the test set. Just wondering what needs to be done here. Should we remove the sorting now or decide once the analysis that you are doing is complete?

badGarnet:
> I have been checking the test errors and I have the impression that the sorting is somehow producing worst outcomes for the tables in the test set. Just wondering what needs to be done here. Should we remove the sorting now or decide once the analysis that you are doing is complete?

I can see why sorting could be a problem: it assumes the table is orientated in certain ways that may not necessarily be true to start with. Let's wait till I look more into the table layout code so to understand how sorting impacts that first before we take an action on this. For now we can let this PR just being stale.

badGarnet:
this was not a good idea

PR #201, opened 2023-09-07
Fix/overlapping of bboxes

benjats07:
This PR implements two major changes:
* Replaces detectron2 with Yolox_quantized as default model
* Introduces an algorithm for reducing nested elements detected by any model. As a benefit of these now is possible to detect more diverse element types.
* Adds a property to `Rectangle` class to register the origin of the data.
* Adds functionality to `annotate` function, to skip elements of certain origins
* Adds functionality to `annotate` function to print additional details of bounding boxes
* Tests updates

The changes made could be tested using the script:
https://gist.github.com/benjats07/dd7b4857a2fbf9ad49a61b609515be9e
(Notice that on the current version, this script doesn't draw elements coming from OCR.)

And the documents (need to be stored at the same location as the script):
https://drive.google.com/drive/folders/1ZsAcraXx5AzfFa9EPjm_IJ6my-2WdJl5?usp=sharing

Is recommended to run this script twice: once with the `unstructured-version` in this PR and again with the version in the `main` branch. After execution, you can check the results in the OUT folder with the following meaning:

[document-number]-[num page]-[model-name]-[old|new].png

where old indicates that the image is generated by the `main` version and `new` indicates output by this PR.

In the -new images you will notice that there aren't any elements inside others.
You can also note that some elements are missing on -new images, this is because all elements not coming from a detection model are being skipped from drawing.


badGarnet:
confirming there is a great reduction in deuplicated boxes but also observing a few interesting differences that might be a regression in detecting certain types of elements (so far found code block and footer/header elements)

image

unstructured_inference/inference/elements.py
```diff
@@ -30,6 +30,7 @@ class Rectangle:
     y1: Union[int, float]
     x2: Union[int, float]
     y2: Union[int, float]
+    source: Optional[str] = None
```
qued, line 33: This would make more sense in one of the child classes like `LayoutElement` or `TextRegion`.

unstructured_inference/inference/layoutelement.py
```diff
@@ -321,13 +327,44 @@ def get_elements_from_ocr_regions(ocr_regions: List[TextRegion]) -> List[LayoutE
...
             r.y2,
+            None,
             text=r.text,
             type="UncategorizedText",
         )
         for r in merged_regions
     ]
 
 
+def probably_contained(
```
qued, line 338: Doesn't this serve the same purpose as `Rectangle.is_almost_subregion_of`?

unstructured_inference/inference/layoutelement.py
```diff
@@ -321,13 +327,44 @@ def get_elements_from_ocr_regions(ocr_regions: List[TextRegion]) -> List[LayoutE
...
+    element: Union[LayoutElement, Rectangle],
+    inside: Union[LayoutElement, Rectangle],
+    area_threshold: float = 0.8,
+) -> bool:
+    """This function checks if one element is inside other.
+        If is definetly inside returns True, in other case check
+    if the intersection is big enough."""
+    # Probably can be implemented usigng is_in, however this is based in area overlapping
```
qued, line 346:
```suggestion
    # Probably can be implemented using is_in, however this is based in area overlapping
```

unstructured_inference/inference/layoutelement.py
```diff
@@ -321,13 +327,44 @@ def get_elements_from_ocr_regions(ocr_regions: List[TextRegion]) -> List[LayoutE
...
+def separate(region_a: Union[LayoutElement, Rectangle], region_b: Union[LayoutElement, Rectangle]):
+    """Reduce leftmost rectangle to don't overlap with the other"""
+    if not region_a.intersects(region_b):
+        return
+    else:
+        leftmost = region_a if region_a.x1 <= region_b.x1 else region_b
+        other = region_b if region_a.x1 <= region_b.x1 else region_a
+
+        leftmost.x2 = other.x1 - 1
+        leftmost.y2 = other.y2 - 1
```
qued, line 365: Won't this create an invalid rectangle if `region_a.x1 == region_b.x1`?

qued, line 365: Changing the x should be enough to ensure the rectangles don't overlap, what's happening with the y coordinate change here? Also it doesn't seem right that they should both be `y2`.

unstructured_inference/models/unstructuredmodel.py
```diff
@@ -55,6 +64,105 @@ def __call__(self, x: Image) -> List[LayoutElement]:
...
+        min_text_size: int = 15,
+    ) -> List[LayoutElement]:
+        """Deletes overlapping elements in a list of elements. Also will delete elements
+        of less than min_text_size pixels height"""
+
+        if len(elements) <= 1:
+            return elements
+
+        def clean_tables(elements: List[LayoutElement]) -> Iterable[LayoutElement]:
+            import numpy as np
```
qued, line 79: Looks like this is a duplicate import.

unstructured_inference/models/unstructuredmodel.py
```diff
@@ -55,6 +64,105 @@ def __call__(self, x: Image) -> List[LayoutElement]:
...
+            return final_elements
+
+        def enhance_regions(
+            elements: List[LayoutElement],
+            min_text_size: int,
+            iou_to_merge: float = 0.5,
+        ) -> List[LayoutElement]:
+            """This function traverse all the elements deleting nested elements,
+            if are close enough could merge it or split them, depending or the
+            iou score for both regions"""
```
qued, line 110:
```suggestion
            """This function traverses all the elements and either deletes nested elements,
            or merges or splits them depending on the iou score for both regions"""
```

unstructured_inference/inference/layoutelement.py
qued: Looking at @badGarnet 's example, I think we need to carefully evaluate whether this method of deduping makes things better or worse a majority of the time.

qued (commented):
I left some initial comments.

unstructured_inference/inference/layoutelement.py
benjats07: I updated this algorithm, but yeah, it created those ones, should be removed because that just happens on elements that are so small that turned to the other side :v

unstructured_inference/inference/layoutelement.py
benjats07: Probably...let me check if I can use that method

unstructured_inference/inference/elements.py
benjats07: Could be, but ...isn't possible that some method creates just Rectangles?

christinestraub: I think `TextRegion` would make more sense.

unstructured_inference/inference/layoutelement.py
christinestraub: It seems like you can use `Rectangle.is_almost_subregion_of`.

From
`nested_current = np.array([probably_contained(e, table) for e in not_tables])`
To
`nested_current = np.array([table.is_almost_subregion_of(e) for e in not_tables])`

unstructured_inference/inference/elements.py
benjats07: Refactoring...

unstructured_inference/inference/layoutelement.py
benjats07: I think the logic is the opposite one, but I'll give it a try. Thanks both 😄

unstructured_inference/models/unstructuredmodel.py
```diff
@@ -55,6 +64,99 @@ def __call__(self, x: Image) -> List[LayoutElement]:
...
+                                # The rectangle could become too small, which is a
+                                # good size to delete?
+                            else:  # big
+                                # merge
+                                if first.area > second.area:
+                                    grow_region_to_match_region(first, second)
+                                    elements[j] = None  # type:ignore
+                                else:
+                                    grow_region_to_match_region(second, first)
+                                    elements[i] = None  # type:ignore
```
badGarnet, line 142: so because this inner loop just goes down the list of all overlapped boxes we could get boxes (lets call it `j+k`) that completely contains this one (`i`) later but another box (current `j`) would have intersection over minimum so we grow `j` while the most appropriate action might be simply dedupe `i` (and later `j` likely would be folded into `j+k` isntead.
It feels like we should first compute the intersection area matrix between pairs of boxes and go from there to merge. That way we know the full picture before changing any elements.

benjats07: Updated to cover that case

unstructured_inference/inference/layoutelement.py
```diff
@@ -322,12 +328,53 @@ def get_elements_from_ocr_regions(ocr_regions: List[TextRegion]) -> List[LayoutE
...
+        # Other is down
+        if reduce.y2 > keep.y2 and reduce.x1 < keep.x2:
+            # other is down-right
+            if reduce.x2 > keep.x2 and reduce.y2 > keep.y2:
+                reduce.x1 = keep.x2 * 1.01
+                reduce.y1 = keep.y2 * 1.01
+                return
+            # other is down-left
+            if reduce.x1 < keep.x1 and reduce.y1 < keep.y2:
+                reduce.x1 = keep.y2
```
badGarnet, line 353: I think you meant to shrink down the `reduce` box in y direction, right?
```suggestion
                reduce.y1 = keep.y2
```

benjats07: Oh, yes :dizzy: 
(By the way, this code isn't called on the last commit, but is good to prepare it anyway)

cragwolfe: follow up question, is there a missing unittest that would have caught that mistake?

unstructured_inference/models/unstructuredmodel.py
```diff
@@ -55,6 +63,116 @@ def __call__(self, x: Image) -> List[LayoutElement]:
...
+        for i, row in enumerate(intersections_mtx):
+            first = elements[i]
+            if first:
+                # We get only the elements which intersected
+                indices_to_check = np.where(row)[0]
+                # Delete the first element, since it will always intersect with itself
+                indices_to_check = indices_to_check[indices_to_check != i]
+                if len(indices_to_check) == 0:
+                    continue
+                if len(indices_to_check) > 1:  # sort by iom
```
badGarnet, line 84: I don't think we need this check since we have the two lines above:? Then we can reduce indentation of the block below

unstructured_inference/models/unstructuredmodel.py
```diff
@@ -55,6 +63,116 @@ def __call__(self, x: Image) -> List[LayoutElement]:
...
+
+        elements = [e for e in elements if e is not None]
+        return elements
+
+    @staticmethod
+    def clean_type(elements: List[LayoutElement], type="Table") -> List[LayoutElement]:
+        """After this function, the list of elements will not contain any element inside
+        of the type specified"""
+        target_elements = [e for e in elements if e.type == "Table"]
+        other_elements = [e for e in elements if e.type != "Table"]
```
badGarnet, line 132: let's also rename this variable to `type_to_clean` so we don't override a reserved word `type`
```suggestion
        target_elements = [e for e in elements if e.type == type_to_clean]
        other_elements = [e for e in elements if e.type != type_to_clean]
```

unstructured_inference/inference/layoutelement.py
benjats07: There is a test but the cases covered aren't enough

unstructured_inference/models/unstructuredmodel.py
```diff
@@ -55,6 +63,116 @@ def __call__(self, x: Image) -> List[LayoutElement]:
...
+        if len(target_elements) == 0:
+            return elements
+
+        nested_check = None
+        for table in target_elements:
```
badGarnet, line 137: lets rename `table` to `element` to avoid confusion (since this function is meant to be able to deal with more than table)

unstructured_inference/models/unstructuredmodel.py
benjats07: You mean `len(index_to_check)>1`, right?
That is needed because will perform unnecessary processing when just have one element

unstructured_inference/models/unstructuredmodel.py
benjats07: OMG, right!

unstructured_inference/models/unstructuredmodel.py
```diff
@@ -55,6 +63,116 @@ def __call__(self, x: Image) -> List[LayoutElement]:
...
+                                elements[j] = None  # type:ignore
+                            else:
+                                grow_region_to_match_region(second, first)
+                                elements[i] = None  # type:ignore
+
+        elements = [e for e in elements if e is not None]
+        return elements
+
+    @staticmethod
+    def clean_type(elements: List[LayoutElement], type="Table") -> List[LayoutElement]:
```
badGarnet: do this new functions added here needs to be in the class? It seems they can just be functions in this sub module and do not need to be part of the class. Especially consider long term reuse maybe it is preferred to have those functions being outside of the class definition?

benjats07: Probably `clean_type` could be outside, but I think `deduplicate_detected_elements` should remain here as only applies to detection models...do you have any particular suggestions on where we can place them?

unstructured_inference/models/unstructuredmodel.py
```diff
@@ -55,6 +63,116 @@ def __call__(self, x: Image) -> List[LayoutElement]:
...
+        elements: List[LayoutElement],
+        iom_to_merge: float = 0.3,
+    ) -> List[LayoutElement]:
+        """This function traverses all the elements and either deletes nested elements,
+        or merges or splits them depending on the iom score for both regions"""
+        intersections_mtx = intersections(*elements)
+
+        for i, row in enumerate(intersections_mtx):
+            first = elements[i]
+            if first:
```
badGarnet, line 77: hmm, there is not chance an element is `None` or at least not pass this test, right? The function `intersections` seem to require all `elements` to be valid and has coord attributes.
Furthermore, if there is a case that this condition returns `False` we'd have undefined `indices_to_check` below at line 95.

unstructured_inference/models/unstructuredmodel.py
```diff
@@ -55,6 +63,116 @@ def __call__(self, x: Image) -> List[LayoutElement]:
...
+                        for j in indices_to_check
+                        if elements[j] is not None
+                    ]
+                    iom_to_check.sort(
+                        key=lambda x: x[1],
+                        reverse=True,
+                    )  # sort elements by iom, so we first check the greatest
+                    indices_to_check = [x[0] for x in iom_to_check]  # type:ignore
+            for j in indices_to_check:
+                if i != j and elements[j] is not None and elements[i] is not None:
```
badGarnet, line 96: lets:
- filter out `i` from `indices_to_check` so this condition is less complicated and we don't have to check for every `j`
- invert this condition to reduce indentation (i.e., invert the condition and `continue` then the rest of the code can be moved one indent forward); this makes the code easier to read

unstructured_inference/models/unstructuredmodel.py
benjats07: Oh, yes, this happens if this element passes through a case when the j=i should be deleted

test_unstructured_inference/models/test_model.py
```diff
@@ -72,3 +72,69 @@ def test_model_initializes_once():
...
+        LayoutElement(0.01, 0.01, 1.01, 1.01),
+        LayoutElement(0.02, 0.02, 1.02, 1.02),
+        LayoutElement(0.03, 0.03, 1.03, 1.03),
+        LayoutElement(0.04, 0.04, 1.04, 1.04),
+        LayoutElement(0.05, 0.05, 1.05, 1.05),
+    ]
+    model = get_model("yolox_tiny")
+    elements = model.enhance_regions(elements, 0.5)
+    assert len(elements) == 1
+    assert (elements[0].x1, elements[0].y1, elements[0].x2, elements[0].x2) == (0, 0, 1.05, 1.05)
```
badGarnet, line 140: this seem to be a copy from the test above?

benjats07: Very similar, but tested the old version when a function only cleaned tables

unstructured_inference/models/unstructuredmodel.py
```diff
@@ -55,6 +63,116 @@ def __call__(self, x: Image) -> List[LayoutElement]:
...
+        elements = [e for e in elements if e is not None]
+        return elements
+
+    @staticmethod
+    def clean_type(elements: List[LayoutElement], type_to_clean="Table") -> List[LayoutElement]:
+        """After this function, the list of elements will not contain any element inside
+        of the type specified"""
+        target_elements = [e for e in elements if e.type == type_to_clean]
+        other_elements = [e for e in elements if e.type != type_to_clean]
+        if len(target_elements) or len(other_elements) == 0:
```
badGarnet, line 133: wait, why would we not want to have any `target_elements`?

benjats07: because... I made a typo.
Should be:
if len(target_elements)==0 or len(other_elements)==0:
If there aren't any of the targets or there aren't any other types to remove, then we return the original list of elements.
Now I'm wondering: what if there are elements of the same type being nested? 🤔

test_unstructured_inference/models/test_model.py
badGarnet: line 138 calls `enhance_regions`

benjats07: Ooops, fixed.
Thanks!

unstructured_inference/models/unstructuredmodel.py
benjats07: I just made the first part, not sure to following you about the second bullet

unstructured_inference/models/unstructuredmodel.py
```diff
@@ -55,6 +63,116 @@ def __call__(self, x: Image) -> List[LayoutElement]:
...
+            nested_current = np.array([e.is_almost_subregion_of(element) for e in other_elements])
+            if nested_check is None:
+                nested_check = nested_current
+                continue
+            nested_check = nested_check | nested_current
+
+        final_elements = []
+        for nested, elem in zip(nested_check, other_elements):  # type:ignore
+            if not nested:
+                final_elements.append(elem)
```
badGarnet, line 147: you can probably use `filter` here to tidy up the code: https://docs.python.org/3/library/functions.html#filter

unstructured_inference/models/unstructuredmodel.py
benjats07: I addressed also the issue of nested types of the same type

unstructured_inference/models/unstructuredmodel.py
benjats07: Done!

unstructured_inference/models/unstructuredmodel.py
benjats07: Solved

unstructured_inference/models/unstructuredmodel.py
badGarnet: I am thinking because those operate on `LayoutElement` so they can be where those are defined. So in this case `unstructure_inference/inference/layoutelements.py`
The idea is that those functions really do not have any idea it is inside a model nor do they need to be. The most relevant context is how `LayoutElement` is defined. So it would seem they are better treated as tools applied to layout elements rather than tools bound to a model (even though right now we only have models using it). Models use the tools just like models depend on layout element's specific definitions. But models doesn't define those tools or layout elements' definitions.
With that said I can let it slide for this ticket and create a followup ticket to refactor these functions out.

test_unstructured_inference/test_elements.py
```diff
@@ -191,6 +195,47 @@ def test_intersection_over_min(
...
+    ("rect1", "rect2"),
+    [
+        (elements.Rectangle(0, 0, 5, 5), elements.Rectangle(3, 3, 6, 6)),
+        (elements.Rectangle(0, 0, 5, 5), elements.Rectangle(6, 6, 8, 8)),
+    ],
+)
+def test_separate(rect1, rect2):
+    separate(rect1, rect2)
+
+    assert not rect1.intersects(rect2)
```
badGarnet, line 236: let's actually also test the new rectangle's coords are what we expected them to be (and modify this test to test all paths in the separate function)

PR #202, opened 2023-09-07
use tesseract info to assist table structure

badGarnet:
## Summary

The table structure post processing code can use OCR software detected block, line, and span order to assist structure detection (by [sorting](https://github.com/Unstructured-IO/unstructured-inference/blob/bdee10268ad056bd3e4313769dcc5a5ef1362155/unstructured_inference/models/table_postprocess.py#L293-L295) the tokens detected by OCR, and this order may not correspond to the order returned by OCR). `tesseract` does provide those information in the table return so this PR adds those data into the `tokens` so they can be utilized in table post processing.

## test

Doesn't seem to change anything yet... this is a small part in the process chain (sorting tokens) so hard to detect if it makes a difference at the moment. More testing might show some difference?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,7 @@
+## 0.5.24
+
+* use `tesseract` extracted block, line, and word order to assist table structure post processing
+
 ## 0.5.23
 
 * Add functionality to bring back embedded images in PDF
--- a/unstructured_inference/__version__.py
+++ b/unstructured_inference/__version__.py
@@ -1,1 +1,1 @@
-__version__ = "0.5.23"  # pragma: no cover
+__version__ = "0.5.24"  # pragma: no cover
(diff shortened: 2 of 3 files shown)


ajjimeno (commented):
The changes should now affect the output except but it seems that outputs are different. It could be due to a change in the model used for table detection (Detectron2, YoloX) or the OCR being used. Should the fixtures be updated if we consider that the output is correct?

badGarnet:
> The changes should now affect the output except but it seems that outputs are different. It could be due to a change in the model used for table detection (Detectron2, YoloX) or the OCR being used. Should the fixtures be updated if we consider that the output is correct?

thanks for checking! yeah I think we should replace the fixture; I am just checking them right now to confirm they are improvements

badGarnet:
> The changes should now affect the output except but it seems that outputs are different. It could be due to a change in the model used for table detection (Detectron2, YoloX) or the OCR being used. Should the fixtures be updated if we consider that the output is correct?

unfortunately the new information actually makes results worse... it seems that it is best to not let OCR provide order information and just trust the table structure model to put them in the right place.

More interestingly in this case throw the whole page into the model (the test itself) works better than going through separated table detection then crop table portion for structure detection (i.e., use `partition_pdf_or_image` from `unstructured`) because yolox and detectron can only detect partial tables.

PR #203, opened 2023-09-08
fix table to html bug

badGarnet:
This PR fixes a bug where tables with cells that spans multiple rows can cause the rows below/aligned with them position incorrectly. The original code assumes the first cell from a row (i.e., the cell that increments the row number) is also the first column. But for tables like the one below (also in this PR as a new sample doc):

table-multi-row-column-cells

The last first 4 columns has headers spanning two rows but the last two columns have headers that has two rows. With the current code the second row (the subheaders) of the last two columns are treated as the subheaders for the first two columns:

Screenshot 2023-09-08 at 11 59 32 AM

This PR fixes this bug by filling in cells with empty text to occupy the cells at row 2, columns 1 through 4.  Those are the cells that are compounded into the header. And the resulting html represents the table structure better:

Screenshot 2023-09-08 at 12 01 26 PM


unstructured_inference/models/tables.py
```diff
@@ -608,9 +608,25 @@ def structure_to_cells(table_structure, tokens):
     return cells, confidence_score
 
 
+def fill_cells(cells: List[dict]) -> List[dict]:
+    """add empty cells to pad cells that spans multiple rows for html conversion
```
cragwolfe, line 612: can you describe the input data structure here which will make this code a bit easier to follow?

unstructured_inference/models/tables.py
```diff
@@ -608,9 +608,25 @@ def structure_to_cells(table_structure, tokens):
...
+            new_cell["cell text"] = ""
+            new_cells.append(new_cell)
+    return new_cells
+
+
 def cells_to_html(cells):
     """Convert table structure to html format."""
-    cells = sorted(cells, key=lambda k: min(k["column_nums"]))
+    cells = sorted(fill_cells(cells), key=lambda k: min(k["column_nums"]))
     cells = sorted(cells, key=lambda k: min(k["row_nums"]))
```
cragwolfe, line 630: this existed before, but confused why there are two distinct sorts

why not:

```
    sorted(fill_cells(cells), key=lambda k: (min(k["row_nums"]), min(k["column_nums"])))
```

qued:
Confirmed the relocation of the column headers to the right spaces. However the outputs are missing some rows: 
image

badGarnet:
> Confirmed the relocation of the column headers to the right spaces. However the outputs are missing some rows: image

this will be in the other PR 😬 https://github.com/Unstructured-IO/unstructured-inference/pull/204

cragwolfe (approved):
🎉

PR #204, opened 2023-09-08
remove cv2 preprocessing

badGarnet:
This PR removes `cv2` preprocessing before OCR step in table transformer.

The preprocessing in practice tend to lead to lower recall from OCR results. For example this image:

table-multi-row-column-cells

with current code:

```python
from unstructured_inference.models.tables import UnstructuredTableTransformerModel
from PIL import Image

model = UnstructuredTableTransformerModel()
model.initialize("microsoft/table-transformer-structure-recognition")
prediction = model.predict(Image.open('table.png'))
```

produces 

Screenshot 2023-09-08 at 9 39 07 AM

after removing the preprocessing (this PR):

Screenshot 2023-09-08 at 12 15 13 PM


cragwolfe (approved):
LGTM pending test change. IMO, we should get this upstream into unstructured asap

PR #205, opened 2023-09-08
add padding before structure detection

badGarnet:
Experiments show that structure detection model works better when padding is added around the image: https://www.notion.so/Investigate-structure-detection-model-9cf53d2aeb6c4a63b44c5324217f7adf

For example, this sample table image (part of this PR) represents a typical input into the structure detection model: a crop of a table that is relatively tightly bounded around the table content.

ilpa-example-1

Without padding (current method), we got extraction html result on this image like:
Screenshot 2023-09-08 at 3 20 49 PM
which misses a few rows that on the top and bottom of the image.

However, once padding is added (this PR, default to 25 pixels on each side) we are able to increase the recall of the structure detection model:
Screenshot 2023-09-08 at 3 22 10 PM

## testing

Please experiment with changing the `pad_for_structure_detection` value and see different results by running

```python
from unstructured_inference.models.tables import UnstructuredTableTransformerModel
from PIL import Image

model = UnstructuredTableTransformerModel()
model.initialize("microsoft/table-transformer-structure-recognition")
prediction = model.predict(Image.open('table.png'))
```

and view the generated html in a browser

## open questions

1. do we want to make this parameter tunable with env variable? are we overdue for a config file?
2. do we want to explore automatically choosing a good padding by experimenting with different images?


ajjimeno:
The changes look reasonable to me. About your questions, I think Crag/Alan are better suited to answer the first question. For the second one, we definitely need to evaluate several values, but just wondering if that should be part of this PR.

qued:
My take for the questions are:
1. I think it will be rare that users will want to tinker with the padding, or dig in enough that they would know how to choose good values, so my take is that this isn't a priority,
2. I think this PR is good without this, but definitely something we want to explore with follow-on work (particularly given my view on 1., it follows that we need to be experts in choosing the value ourselves).

unstructured_inference/models/tables.py
```diff
@@ -100,11 +101,16 @@ def get_tokens(self, x: Image):
             )
         return tokens
 
-    def run_prediction(self, x: Image):
+    def run_prediction(self, x: Image, pad_for_structure_detection: int = 50):
         """Predict table structure"""
         with torch.no_grad():
-            encoding = self.feature_extractor(x, return_tensors="pt").to(self.device)
+            logger.info(f"padding image by {pad_for_structure_detection} for structufre detection")
```
qued, line 107:
```suggestion
            logger.info(f"padding image by {pad_for_structure_detection} for structure detection")
```

unstructured_inference/utils.py
```diff
@@ -123,3 +123,22 @@ def annotate_layout_elements(
...
+
+
+def pad_image_with_background_color(
+    image: Image.Image,
+    pad: int = 10,
+    background_color: str = "white",
+) -> Image.Image:
+    """pads an input image with the same background color around it by pad//2 on all 4 sides
+
+    The original image is kept inact and a new image is returned with padding added.
```
qued, line 135:
```suggestion
    The original image is kept intact and a new image is returned with padding added.
```

qued (approved):
LGTM, just correct two typos in the documentation.

Issue #200, opened 2023-09-07
Use PaddleOCR if platform.machine returns AMD64 (or other potentially compatible platform)


qued:
Currently PaddleOCR is used only if we detect the `x86_64` platform. There are potentially other platforms that are compatible with PaddleOCR, including `AMD64`. This issue is to investigate which platforms we should support for PaddleOCR, and update the code to reflect that.

PR #207, opened 2023-09-11
chore: support paddle with both cpu and gpu if it is installed

yuming-long:
address: https://github.com/Unstructured-IO/unstructured-inference/issues/200

### Summary
* Before this PR, we only used paddle if the platform is `x86_64`, remove this platform check and just assume that paddle and paddle OCR is already installed on user's platform
* Introduce an env variable `TABLE_OCR` that can be used to specify if we use paddle/tesseract for table extraction, default is tesseract
* Add logic for parameter `use_gpu` while setting up paddle ocr instance, if there is gpu advices available, paddle will use gpu

### Test
* paddle on CPU: please check unit test `test_table_prediction_paddle`, the result is not good (can use gdb to print value of `prediction`) but can prove that paddle ocr is running in CI tests
* paddle on GPU: on a gpu instance:
   * rerun test `PYTHONPATH=. pytest test_unstructured_inference/models/test_tables.py::test_table_prediction_paddle` and run 
   * from another terminal run `nvidia-smi -l 1` to monitor gpu usage, and you can see the python test is using gpu memory


jasonbot (approved):

unstructured_inference/models/paddle_ocr.py
```diff
@@ -1,12 +1,38 @@
+import paddle
+from unstructured_paddleocr import PaddleOCR
+
 paddle_ocr = None  # type: ignore
```
jasonbot, line 4: Rather than do this `global` dance, it'd be more explicit to just `@functools.cache`the `load_agent` function

ajjimeno:
LGTM! I testing it on an instance with GPU. nvidia-smi showed that the memory was being used.

PR #208, opened 2023-09-11
Feat/save embedded images in pdf

christinestraub:
Addresses unstructured issue [#1332](https://github.com/Unstructured-IO/unstructured/issues/1332). This PR will work together with unstructured PR [#1371](https://github.com/Unstructured-IO/unstructured/pull/1371).
This PR also addresses `"true" embedded images` issue #215.

### Summary
- Add functionality to extract and save images from the page
  - add the `extract_images` method to the `PageLayout` class
  - pass parameters related to extracting images from the page
  - add Python script to evaluate image extraction with various PDF processing libraries
- Add functionality to get only "true" embedded images when extracting elements from PDF pages
  - add functionality to extract image objects (`LTImage`) from a `PDF layout element` parsed by `pdfminer.high_level.extract_pages`
  - update logic to determine `ImageTextRegion` in `load_pdf()`
- Update the `layout visualization` script to be able to show only image elements if need

The following documents can be used for testing and evaluation.
- [Captur-1317-5_ENG-p23.pdf](https://utic-dev-tech-fixtures.s3.us-east-2.amazonaws.com/pastebin/Captur-1317-5_ENG-p23.pdf)
- [23-BERKSHIRE.pdf](https://utic-dev-tech-fixtures.s3.us-east-2.amazonaws.com/pastebin/23-BERKSHIRE.pdf)
- [main.PMC6312790-p1.pdf](https://github.com/Unstructured-IO/unstructured-inference/files/12675967/main.PMC6312790_1-1.pdf)

### Testing
```

f_path = "sample-docs/embedded-images.pdf"

# default image output directory
    filename=f_path,
    extract_images_in_pdf=True,
)

# specific image output directory
    filename=f_path,
    extract_images_in_pdf=True,
    image_output_dir_path=<directory_path>,
)
```
### Evaluation
```
// Extracting Images
$ PYTHONPATH=. python examples/image-extraction/embedded-image-extraction.py Captur-1317-5_ENG-p23.pdf unstructured

// Layout Visualziation
$ PYTHONPATH=. python examples/layout_analysis/visualization.py Captur-1317-5_ENG-p23.pdf image_oly
```
**NOTE:** To reproduce the original results for comparision, you need to replace [the lines](https://github.com/Unstructured-IO/unstructured-inference/blob/feat/save-embedded-images-in-pdf/unstructured_inference/inference/layout.py#L650-L659) with the following code snippet
```
_text, element_class = (
    (element.get_text(), EmbeddedTextRegion)
    if hasattr(element, "get_text")
    else (None, ImageTextRegion)
)
```


unstructured_inference/inference/layout.py
```diff
@@ -427,6 +455,8 @@ def from_image(
         """Creates a PageLayout from an already-loaded PIL Image."""
         analysis = kwargs.get("analysis", False)
         supplement_with_ocr_elements = kwargs.get("supplement_with_ocr_elements", True)
+        extract_images_in_pdf = kwargs.get("extract_images_in_pdf", False)
+        image_output_dir_path = kwargs.get("image_output_dir_path", None)
```
qued, line 459: You didn't start this pattern, but why don't we just move all these to the signature.

unstructured_inference/inference/layout.py
```diff
@@ -339,6 +340,33 @@ def get_elements_from_layout(self, layout: List[TextRegion]) -> List[LayoutEleme
...
+            figure_number += 1
+            try:
+                output_f_path = os.path.join(
+                    output_dir_path,
+                    f"figure-{self.number}-{figure_number}.jpg",
+                )
+                cropped_image = self.image.crop((el.x1, el.y1, el.x2, el.y2))
+                write_image(cropped_image, output_f_path)
+                el.image_path = output_f_path
+            except Exception:
```
qued, line 367: Are there specific errors this is in place to account for? I'd prefer not to use such a general exception class here.

qued (commented):
Overall looks good, just a pair of small change requests. Meanwhile I haven't run the test code yet, so that's my next step.

unstructured_inference/inference/layout.py
qued: By "all these" I mean `analysis`, `supplement_with_ocr_elements`, `extract_images_in_pdf`, and `image_output_dir_path`.

qued (changes requested):
Several things on further inspection / after running test code:
1. If I understand the motivation behind this issue correctly, I think we want the _embedded_ image, meaning the image that is actually stored in the PDF. This appears to identify the image location, then use the bounding box information to crop that portion of the rendered PDF and save it. This results in something potentially different in format and resolution than what the PDF stores internally. (@cragwolfe correct me if I'm off base here)
2. Running the test code results in a few things that seem off to me. It produces outputs `figure-1-x.jpg` where `x` ranges from 1 to 14, skipping 5. I get the logged error message showing that the exception handler triggers (presumably on image 5). Is it expected that we get an error on this sample doc? 
3. Additionally output images 3 and 4 appear to be duplicates, 6 and 7 appear to be duplicates, 8 and 9 appear to be duplicates, 10 and 11 appear to be duplicates, and 13 and 14 appear to be duplicates.

christinestraub:
> Several things on further inspection / after running test code:
> 
> 1. If I understand the motivation behind this issue correctly, I think we want the _embedded_ image, meaning the image that is actually stored in the PDF. This appears to identify the image location, then use the bounding box information to crop that portion of the rendered PDF and save it. This results in something potentially different in format and resolution than what the PDF stores internally. (@cragwolfe correct me if I'm off base here)
> 2. Running the test code results in a few things that seem off to me. It produces outputs `figure-1-x.jpg` where `x` ranges from 1 to 14, skipping 5. I get the logged error message showing that the exception handler triggers (presumably on image 5). Is it expected that we get an error on this sample doc?
> 3. Additionally output images 3 and 4 appear to be duplicates, 6 and 7 appear to be duplicates, 8 and 9 appear to be duplicates, 10 and 11 appear to be duplicates, and 13 and 14 appear to be duplicates.

1. The resolution (width, height) of the rendered PDF page image depends on the `dpi`(https://github.com/Unstructured-IO/unstructured-inference/blob/main/unstructured_inference/inference/layout.py#L612-L624) and the bounding boxes of the image elements are also calculated using the same `dpi`(https://github.com/Unstructured-IO/unstructured-inference/blob/main/unstructured_inference/inference/layout.py#L597-L603) so I don't think there would be potentially different in format and resolution than what the PDF stores internally.
2. Yes, It is expected that we get an error on this sample doc. The width of the cropped image is `0` for the `image 5` (figure-1-5.jpg) so it gets the error - `ValueError: cannot write empty image as JPEG`.
3. It is the issue of the `pdfminer` library (`pdfminer.high_level.extract_pages`). @crag and I talked about that issue in [this Slack channel](https://unstructuredw-kbe4326.slack.com/archives/C057R3F8F7A/p1694538162701109?thread_ts=1694497553.786149&cid=C057R3F8F7A).

unstructured_inference/inference/layout.py
christinestraub: Sounds good!

christinestraub:
@cragwolfe @qued Reverted back to the former implementation (extracting embedded images by cropping each page image based on the image elements bounding boxes) with some improvements.

christinestraub:
As of now, some extracted image elements are missing in the final result and are not being saved. This issue will be addressed by issue #219.

cragwolfe (approved):

PR #209, opened 2023-09-14
Chore: add env `ENTIRE_PAGE_OCR` to specify paddle/tesseract for entire page ocr

yuming-long:
### Summary

We need a way to use paddle for the entire page OCR since the OCR result could be better than tesseract, which has shown on some image files with tables. This PR adds an environment variable `ENTIRE_PAGE_OCR` that can be set to `paddle` or `tesseract`. We still use tesseract as default since paddle performs poorly on entire-page English PDF files.

### Test
if you are on x86 arch, please run this snippet to install paddle (paddle still doesn't work on m1/m2 chip locally):
```
pip install paddlepaddle #or pip install unstructured.paddlepaddle if on aarch64 arch
pip install unstructured_paddleocr
export LD_LIBRARY_PATH=/usr/local/lib:/usr/local/lib64
```
run the following script to see different entire page result from paddle and tesseract
```
import os 

def get_layout_from_image(ocr_languages):
    layout = DocumentLayout.from_image_file("sample-docs/table-multi-row-column-cells.png", ocr_languages=ocr_languages)
    # Create a list to store the layout elements with only "text" and "type" fields
    elements_dict_list = []
        
    for page in layout.pages:
        for element in page.elements:
            element_dict = {
                "text": element.text,
                "type": element.type
            }
            elements_dict_list.append(element_dict)
    return elements_dict_list


# default is tesseract
os.environ['ENTIRE_PAGE_OCR'] = "tesseract"
tesseract_elements = get_layout_from_image(ocr_languages="eng")

# set env to use paddle and call function agin
os.environ['ENTIRE_PAGE_OCR'] = "paddle"
paddle_elements = get_layout_from_image(ocr_languages="en")

# should expect difference
assert tesseract_elements != paddle_elements
# compare result
print(tesseract_elements)
print(paddle_elements)
```

### Note
There are different language code between tesseract and paddle on the same language i.e, `en` in paddle and `eng` in tesseract for English. This can be addressed once we introduce the language mappings from standard language code to tesseract and to paddle respectively. However, unlike tesseract, paddle does support passing in multiple languages, and we will fallback to tesseract if thats the case (future PR).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,7 @@
+## 0.5.28
+
+* add env variable `ENTIRE_PAGE_OCR` to specify using paddle or tesseract on entire page OCR
+
 ## 0.5.27
 
 * table structure detection now pads the input image by 25 pixels in all 4 directions to improve its recall
--- a/test_unstructured_inference/inference/test_layout.py
+++ b/test_unstructured_inference/inference/test_layout.py
@@ -157,8 +157,42 @@ class MockPool:
     def join(self):
         pass
 
+@pytest.mark.parametrize("entire_page_ocr", ["paddle", "tesseract"])
+def test_get_page_elements_with_ocr(monkeypatch, entire_page_ocr):
+    monkeypatch.setenv("ENTIRE_PAGE_OCR", entire_page_ocr)
+    text_block = layout.TextRegion(2, 4, 6, 8, text=None)
+    image_block = layout.ImageTextRegion(8, 14, 16, 18)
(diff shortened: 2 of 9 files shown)


shreyanid:
Comment on the note on language mapping: tesseract does also support passing in multiple languages in a single string, using a plus sign to concatenate the codes (ex. "eng+kor").

christinestraub:
LGTM. After running the test code, I can see paddle performs poorly on this sample doc.

christinestraub (approved):

PR #211, opened 2023-09-15
Fix enhance/duplicated bboxes

benjats07:
Similar changes and ways to test than:
https://github.com/Unstructured-IO/unstructured-inference/pull/201

But applies the deduplication with all the elements (not just coming from detection models)


unstructured_inference/inference/layout.py
```diff
@@ -371,6 +373,8 @@ def annotate(
         colors: Optional[Union[List[str], str]] = None,
         image_dpi: int = 200,
         annotation_data: Optional[dict[str, dict]] = None,
+        add_details: bool = False,
+        sources: List[str] = ["all"],
```
badGarnet, line 377: probably wroth some docstring to document this behavior

unstructured_inference/inference/layoutelement.py
```diff
@@ -59,7 +59,9 @@ def from_region(cls, region: Rectangle):
         x1, y1, x2, y2 = region.x1, region.y1, region.x2, region.y2
         text = region.text if hasattr(region, "text") else None
         type = region.type if hasattr(region, "type") else None
-        return cls(x1, y1, x2, y2, text, type)
+        prob = region.prob if hasattr(region, "prob") else None
```
badGarnet, line 62:
```suggestion
        prob = getattr(region, "prob")
```
and similar below

unstructured_inference/inference/layoutelement.py
```diff
@@ -68,11 +70,13 @@ def from_lp_textblock(cls, textblock: TextBlock):
         text = textblock.text
         type = textblock.type
         score = textblock.score
-        return cls(x1, y1, x2, y2, text, type, prob=score)
+        return cls(x1, y1, x2, y2, "detectron2_lp", text, type, score)
```
badGarnet, line 73: lets either put `"detectron2_lp" into a kwarg or set as a constant somewhere so it is clear to caller this is what will be used.

unstructured_inference/inference/layoutelement.py
```diff
@@ -301,8 +306,9 @@ def merge_text_regions(regions: List[TextRegion]) -> TextRegion:
     max_y2 = max([tr.y2 for tr in regions])
 
     merged_text = " ".join([tr.text for tr in regions if tr.text])
-
-    return TextRegion(min_x1, min_y1, max_x2, max_y2, merged_text)
+    sources = [*{tr.source for tr in regions}]
```
badGarnet, line 309: hmm, is this first get a set then convert to list? maybe `list({tr.source for tr in regions})` would be clearer?

unstructured_inference/inference/layoutelement.py
```diff
@@ -301,8 +306,9 @@ def merge_text_regions(regions: List[TextRegion]) -> TextRegion:
...
+    source = sources.pop() if len(sources) == 1 else "merged:".join(sources)  # type:ignore
```
badGarnet, line 310: the string "merged" should probably be part of the class attribute for `TextRegion` or defined somewhere that can be imported so that downstream can know what joiner was used to join the list. Also you can just say `"merged:".join(sources)`. If the list has only one member it will just return itself, the same as `souces.pop()`

unstructured_inference/inference/layoutelement.py
```diff
@@ -321,13 +327,72 @@ def get_elements_from_ocr_regions(ocr_regions: List[TextRegion]) -> List[LayoutE
...
+        If is definetly inside returns True, in other case check
+    if the intersection is big enough."""
+    # Probably can be implemented usigng is_in, however this is based in area overlapping
+    if element.is_in(inside):
+        return True
+
+    intersected_area = element.intersection(inside)
+    if intersected_area:
+        return intersected_area.area >= inside.area * area_threshold
+    return False
```
badGarnet, line 353: This is the core, well one of, of this PR it seems.
area overlap seem ok but may have recall issues: e.g., the boxes "Word", "oil" are ok to be not merged (if they were to be merged we might have other issues. But the box "Emerging" seem like a case it should be because the box bounding the text has high overlap on width and height: the point is area is width * height so suppose both sides got 0.9 overlap but the area would be only 0.81 overlap at best. 
Screenshot 2023-09-15 at 10 47 58 AM

unstructured_inference/inference/layoutelement.py
```diff
@@ -321,13 +327,72 @@ def get_elements_from_ocr_regions(ocr_regions: List[TextRegion]) -> List[LayoutE
...
+
+    def reduce(keep: Rectangle, reduce: Rectangle):
+        # Asume intersection
+
+        # Other is down
+        if reduce.y2 > keep.y2 and reduce.x1 < keep.x2:
+            # other is down-right
+            if reduce.x2 > keep.x2 and reduce.y2 > keep.y2:
+                reduce.x1 = keep.x2 * 1.01
+                reduce.y1 = keep.y2 * 1.01
```
badGarnet, line 367: let's define constant values like `1.01` here somewhere that can be imported / changed.

unstructured_inference/models/unstructuredmodel.py
```diff
@@ -55,6 +64,106 @@ def __call__(self, x: Image) -> List[LayoutElement]:
...
+        elements: List[LayoutElement],
+        min_text_size: int = 15,
+    ) -> List[LayoutElement]:
+        """Deletes overlapping elements in a list of elements. Also will delete elements
+        of less than min_text_size pixels height"""
+
+        if len(elements) <= 1:
+            return elements
+
+        def clean_tables(elements: List[LayoutElement]) -> Iterable[LayoutElement]:
```
badGarnet, line 78: this and the function below `enhance_regions` are important unit steps in the code logic so it would be better to refactor them out of this parent function scope so that:
- logic is isolated
- can test cleanly the unit steps instead of having to test dedupe as a whole only

unstructured_inference/models/unstructuredmodel.py
```diff
@@ -55,6 +64,106 @@ def __call__(self, x: Image) -> List[LayoutElement]:
...
+                first = elements[i]
+                if first:
+                    # We get only the elements which intersected
+                    indices_to_check = np.where(row)[0]
+                for j in indices_to_check:
+                    if i != j and elements[j] is not None and elements[i] is not None:
+                        second = elements[j]
+                        intersection = first.intersection(
+                            second,
+                        )  # we know it does, but need the region
```
badGarnet, line 123: so we have computed somewhere -> can we capture that when it was first computed so here we just use the reference instead of recomputing it here?

benjats07: Not currently, but is an issue that could be addressed to get performance improvement 🤔

unstructured_inference/inference/layoutelement.py
benjats07: Actually, think this requires more than that. Until now I've been hardcoding the name on code just called with certain models, but probably we need all our models to contain a "model_name" variable for simplicity.

unstructured_inference/inference/layoutelement.py
benjats07: Actually, after the refactor, this never happens, because Rectangle doesn't contain those properties. Nice catch anyway 😄

PR #212, opened 2023-09-15
Enhance/duplicated bboxes all document

benjats07:
Similar changes and ways to test than:
https://github.com/Unstructured-IO/unstructured-inference/pull/201

But applies the deduplication with all the elements (not just coming from detection models)


unstructured_inference/inference/elements.py
```diff
@@ -118,27 +119,27 @@ def intersection_over_union(self, other: Rectangle) -> float:
         how similar the regions are. Returns 0 for disjoint rectangles, 1 for two identical
         rectangles -- area of intersection / area of union."""
         intersection = self.intersection(other)
-        intersection_area = 0.0 if intersection is None else intersection.area()
-        union_area = self.area() + other.area() - intersection_area
+        intersection_area = 0.0 if intersection is None else intersection.area
```
badGarnet, line 122: since `intersection_area` is used more than once it is a sign we should refactor it as its own helper function

PR #218, opened 2023-09-20
feat: add config class

badGarnet:
- add a dataclass that contains configrations for inference processes
- the parameters can be specified via env variables, which overrides the default values; this allows for flexibly setup in different applications/deployments
- currently contains specifications for table and layout related parameters
- followup needed to identify other parameters that can be added to this config class

### DIFF
--- a/unstructured_inference/config.py
+++ b/unstructured_inference/config.py
@@ -1,0 +1,75 @@
+import os
+from dataclasses import dataclass
+from typing import Optional
+
+
+@dataclass
+class InferenceConfig:
+    """class for configuring inference parameters"""
+
+    def _get_string(self, var: str, default_value: Optional[str] = None) -> str:
+        """attempt to get the value of var from the os environment; if not present return the
+        default_value"""
+        return os.environ.get(var, default_value)
+
+    def _get_int(self, var: str, default_value: int) -> int:
+        if (value := self._get_string(var)) is not None:
+            return int(value)
+        return default_value
+
+    def _get_float(self, var: str, default_value: float) -> float:
+        if (value := self._get_string(var)) is not None:
+            return float(value)
+        return default_value
+
+    @property
+    def TABLE_IMAGE_CROP_PAD(self) -> int:
+        """extra image content to add around an identified table region; measured in pixels
+
+        The padding adds image data around an identified table bounding box for downstream table
+        structure detection model use as input
+        """
+        return self._get_int("TABLE_IMAGE_CROP_PAD", 12)
+
+    @property
+    def TABLE_IMAGE_BACKGROUN_PAD(self) -> int:
+        """number of pixels to pad around an table image with a white background color
+
+        The padding adds NO image data around an identified table bounding box; it simply adds white
+        background around the image
+        """
+        return self._get_int("TABLE_IMAGE_BACKGROUN_PAD", 0)
+
+    @property
+    def LAYOUT_SAME_REGION_THRESHOLD(self) -> float:
+        """threshold for two layouts' bounding boxes to be considered as the same region
+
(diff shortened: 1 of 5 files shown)


unstructured_inference/config.py
```diff
@@ -0,0 +1,75 @@
...
+    def TABLE_IMAGE_CROP_PAD(self) -> int:
+        """extra image content to add around an identified table region; measured in pixels
+
+        The padding adds image data around an identified table bounding box for downstream table
+        structure detection model use as input
+        """
+        return self._get_int("TABLE_IMAGE_CROP_PAD", 12)
+
+    @property
+    def TABLE_IMAGE_BACKGROUN_PAD(self) -> int:
```
qued, line 34:
```suggestion
    def TABLE_IMAGE_BACKGROUND_PAD(self) -> int:
```

unstructured_inference/config.py
```diff
@@ -0,0 +1,75 @@
...
+        """number of pixels to pad around an table image with a white background color
+
+        The padding adds NO image data around an identified table bounding box; it simply adds white
+        background around the image
+        """
+        return self._get_int("TABLE_IMAGE_BACKGROUN_PAD", 0)
```
qued, line 40:
```suggestion
        return self._get_int("TABLE_IMAGE_BACKGROUND_PAD", 0)
```

unstructured_inference/models/tables.py
```diff
@@ -113,7 +114,11 @@ def get_tokens(self, x: Image):
 
         return tokens
 
-    def get_structure(self, x: Image, pad_for_structure_detection: int = 50) -> dict:
+    def get_structure(
+        self,
+        x: Image,
+        pad_for_structure_detection: int = inference_config.TABLE_IMAGE_BACKGROUN_PAD,
```
qued, line 120:
```suggestion
        pad_for_structure_detection: int = inference_config.TABLE_IMAGE_BACKGROUND_PAD,
```

unstructured_inference/models/tables.py
```diff
@@ -126,7 +131,11 @@ def get_structure(self, x: Image, pad_for_structure_detection: int = 50) -> dict
             outputs_structure["pad_for_structure_detection"] = pad_for_structure_detection
             return outputs_structure
 
-    def run_prediction(self, x: Image, pad_for_structure_detection: int = 50):
+    def run_prediction(
+        self,
+        x: Image,
+        pad_for_structure_detection: int = inference_config.TABLE_IMAGE_BACKGROUN_PAD,
```
qued, line 137:
```suggestion
        pad_for_structure_detection: int = inference_config.TABLE_IMAGE_BACKGROUND_PAD,
```

unstructured_inference/inference/layoutelement.py
```diff
@@ -8,6 +8,7 @@
 from pandas import DataFrame
 from PIL import Image
 
+from unstructured_inference.config import inference_config
 from unstructured_inference.constants import FULL_PAGE_REGION_THRESHOLD, SUBREGION_THRESHOLD_FOR_OCR
```
qued, line 12: Let me make sure I understand the convention... the constants in `unstructured_inference.constants` don't belong in `inference_config` because we don't want them to be overridden by environment variables. Is that correct?

qued (approved):
LGTM, just a few typos and a request for clarification.

unstructured_inference/inference/layoutelement.py
badGarnet: right; constants are variables that are NOT configurable. There is just one value we will use. While config contains values we want to tweak and may even need to change for different applications.
Let me add a comment to the module file to clarify

PR #224, opened 2023-09-21
feat: make table transformer parameters configurable

badGarnet:
- refactor `tables.py` so that the structure element confidence threshold values are loaded from `inference_config`
- refactor intersection over box area threshold in `objects_to_structure` to config intead of using hardwired value of 0.5 (default is still 0.5)


qued (approved):
LGTM!

PR #226, opened 2023-09-22
chore: stop passing language code from tesseract mapping to paddle

yuming-long:
### Summary

A user is flagging the assertion error for paddle language code:
```
AssertionError: param lang must in dict_keys(['ch', 'en', 'korean', 'japan', 'chinese_cht', 'ta', 'te', 'ka', 'latin', 'arabic', 'cyrillic', 'devanagari']), but got eng
```
and tried setting the `ocr_languages` param to 'en' (the correct lang code for english in paddle) but also didn't work.
The reason is that the `ocr_languages` uses the mapping for tesseract code which will convert `en` to `eng` since thats the correct lang code for english in tesseract.

The quick workaround here is stop passing the lang code to paddle and let it use default `en`, and this will be addressed once we have the lang code mapping for paddle.

### Test
looks like user used this branch and got the lang parameter working from [linked comments](https://github.com/Unstructured-IO/unstructured-api/issues/247#issuecomment-1731923667) :)
on api repo:
```
pip install paddlepaddle
pip install "unstructured.PaddleOCR"
export ENTIRE_PAGE_OCR=paddle
make run-web-app
```
* check error before this change:
```
curl  -X 'POST'  'http://localhost:8000/general/v0/general'   -H 'accept: application/json'  -F 'files=@sample-docs/english-and-korean.png'   -F 'ocr_languages=en'  | jq -C . | less -R
```
will see the error:
```
{
  "detail": "param lang must in dict_keys(['ch', 'en', 'korean', 'japan', 'chinese_cht', 'ta', 'te', 'ka', 'latin', 'arabic', 'cyrillic', 'devanagari']), but got eng"
}
```
also in logger you will see `INFO Loading paddle with CPU on language=eng...` since tesseract mapping converts `en` to `eng`.
* check after this change:

Checkout to this branch and install inference repo into your env (the same env thats running api) with `pip install -e .`

Rerun `make run-web-app`

Run the curl command again, you won't get the result on m1 chip since paddle doesn't work on it but from the logger info you can see `2023-09-27 12:48:48,120 unstructured_inference INFO Loading paddle with CPU on language=en...`, which means the lang parameter is using default `en` (logger info is coming from [this line](https://github.com/Unstructured-IO/unstructured-inference/blob/main/unstructured_inference/models/paddle_ocr.py#L22)).


yuming-long:
thanks! just added more test description

unstructured_inference/inference/layout.py
```diff
@@ -278,9 +278,8 @@ def get_elements_with_detection_model(
                 logger.info("Processing entrie page OCR with paddle...")
                 from unstructured_inference.models import paddle_ocr
 
-                # TODO(yuming): paddle only support one language at once,
-                # change ocr to tesseract if passed in multilanguages.
-                ocr_data = paddle_ocr.load_agent(language=self.ocr_languages).ocr(
+                # TODO(yuming): pass ocr language to paddle when we have laguage mapping for paddle
```
shreyanid, line 281:
```suggestion
                # TODO(yuming): pass ocr language to paddle when we have language mapping for paddle
```
nit

unstructured_inference/inference/layout.py
```diff
@@ -278,9 +278,8 @@ def get_elements_with_detection_model(
                 logger.info("Processing entrie page OCR with paddle...")
```
shreyanid, line 278:
```suggestion
                logger.info("Processing entire page OCR with paddle...")
```
nit

shreyanid (approved):
lgtm other than some spelling nits, test worked for me! I also tested with ocr_languages="eng" and was able to reproduce the correct `INFO Loading paddle with CPU on language=en..` message indicating that it was using the default paddle value. thanks for the fix!

yuming-long:
@shreyanid do you have access to force merge this PR (since ingest tests are doomed to fail lol

shreyanid:
I guess I do :)

PR #231, opened 2023-09-29
Refactor: Remove OCR related code for entire page OCR

yuming-long:
## Summary
One part of OCR refactor to move it from inference repo to unstructured repo. This PR removes all OCR related code for entire page OCR, which means all table related OCR still remain the same (will be moved after table refactor to accept preprocessed OCR data) 

## Test
Please see test description in https://github.com/Unstructured-IO/unstructured/pull/1579, since those two need to work together.

## Note
The ingest test won't pass until we merge the unstructured refactor PR


unstructured_inference/inference/layout.py
```diff
@@ -261,40 +229,6 @@ def get_elements_with_detection_model(
         # NOTE(mrobinson) - We'll want make this model inference step some kind of
         # remote call in the future.
         inferred_layout: List[LayoutElement] = self.detection_model(self.image)
-        inferred_layout = UnstructuredObjectDetectionModel.deduplicate_detected_elements(
```
benjats07, line 264: This call to deduplicate_detected_elements is required to handle nested elements generated by detection models

benjats07 (changes requested):

unstructured_inference/inference/layout.py
yuming-long: oops i deleted too much

benjats07 (approved):

Dockerfile
```diff
@@ -20,7 +20,6 @@ RUN python3.8 -m pip install pip==${PIP_VERSION} && \
   pip install --no-cache -r requirements/base.txt && \
   pip install --no-cache -r requirements/test.txt && \
   pip install --no-cache -r requirements/dev.txt && \
-  pip install "unstructured.PaddleOCR" && \
```
christinestraub, line 23: I guess we still use the paddle_ocr for tables?

unstructured_inference/inference/elements.py
```diff
@@ -208,18 +205,15 @@ def extract_text(
...
-        ocr_strategy: str = "auto",
-        ocr_languages: str = "eng",
     ) -> str:
         """Extracts text contained in region."""
         if self.text is not None:
             # If block text is already populated, we'll assume it's correct
             text = self.text
         elif objects is not None:
-            text = aggregate_by_block(self, image, objects, ocr_strategy)
+            text = aggregate_by_block(self, image, objects)
```
christinestraub, line 214:
```suggestion
            text = aggregate_by_block(self, objects)
```

unstructured_inference/inference/elements.py
```diff
@@ -208,18 +205,15 @@ def extract_text(
...
         elif image is not None:
-            # We don't have anything to go on but the image itself, so we use OCR
-            text = ocr(self, image, languages=ocr_languages) if ocr_strategy != "never" else ""
+            text = ""
```
christinestraub, line 216: These lines can be removed.

unstructured_inference/inference/elements.py
```diff
@@ -208,18 +205,15 @@ def extract_text(
         objects: Optional[Collection[TextRegion]],
         image: Optional[Image.Image] = None,
```
christinestraub, line 206: The `image` parameter can be removed.

unstructured_inference/inference/elements.py
```diff
@@ -208,18 +205,15 @@ def extract_text(
...
             text = self.text
         elif objects is not None:
-            text = aggregate_by_block(self, image, objects, ocr_strategy)
+            text = aggregate_by_block(self, image, objects)
         elif image is not None:
-            # We don't have anything to go on but the image itself, so we use OCR
-            text = ocr(self, image, languages=ocr_languages) if ocr_strategy != "never" else ""
+            text = ""
         else:
             raise ValueError(
```
christinestraub, line 218:
```suggestion
            text = ""
```

unstructured_inference/inference/elements.py
```diff
@@ -234,8 +228,6 @@ def extract_text(
         objects: Optional[Collection[TextRegion]],
         image: Optional[Image.Image] = None,
```
christinestraub, line 229: The `image` parameter can be removed.

unstructured_inference/inference/elements.py
```diff
@@ -250,96 +242,23 @@ def extract_text(
...
```
christinestraub, line 243: The `image` parameter can be removed.

unstructured_inference/inference/elements.py
```diff
@@ -250,96 +242,23 @@ def extract_text(
...
-                return False
-            else:
-                return image_intersects
         else:
-            # If the region has text, we should only have to OCR if too much of the text is
-            # uninterpretable.
-            return cid_ratio(region.text) > 0.5
-    else:
-        return False
+            return super().extract_text(objects, image, extract_tables)
```
christinestraub, line 250:
```suggestion
            return super().extract_text(objects, extract_tables)
```

christinestraub (changes requested):

Dockerfile
yuming-long: yep we do but maybe no need to keep it in dockerfile
1. we updated the forked library "unstructured.PaddleOCR" so that it doesn't use `paddlepaddle` as a dependency, so this paddle install here won't work since its missing a install for `paddlepaddle` ([updated way to install paddle](https://github.com/Unstructured-IO/unstructured/blob/main/scripts/install-paddleocr.sh)).
2. i don't think there are users using the inference dockerfile, and the install in docker will be moved to unstructured once we are ready to add it

CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+## 0.6.7
```
cragwolfe, line 1:
```suggestion
## 0.7.0
```
significant change, so let's bump to 0.7.0

unstructured_inference/inference/elements.py
yuming-long: i can remove how image is used in extract_text but seems like i still have to pass it to the function (i can pass a None to it):
in
```
@dataclass
class LayoutElement(TextRegion):
    type: Optional[str] = None
    prob: Optional[float] = None
    image_path: Optional[str] = None

    def extract_text(
        self,
        objects: Optional[Collection[TextRegion]],
        image: Optional[Image.Image] = None,
        extract_tables: bool = False,
    ):
        """Extracts text contained in region"""
        text = super().extract_text(
            objects=objects,
            image=image,
            extract_tables=extract_tables,
        )
        if extract_tables and self.type == "Table":
            self.text_as_html = interpret_table_block(self, image)
        return text
   ```
   the child class  `LayoutElement` is using image in `extract_text` and signature of `extract_text` in child class `LayoutElement` and parent class `TextRegion` should be the same, otherwise i see `make check` error: `Signature of "extract_text" incompatible with supertype "TextRegion"  [override]`

File: unstructured_inference/inference/elements.py (hunk 6 above)
Comment (yuming-long) [reply unstructured_inference/inference/elements.py:216]: i removed the whole block of `elif image is not None:` and change the value error

Comment (christinestraub) [reply unstructured_inference/inference/elements.py:216]: @yuming-long IMO, the `else` clause should be set to `text = ""` without raising a `ValueError`. Otherwise, image file processing will fail. 

```
    f_path = "sample-docs/layout-parser-paper-fast.jpg"

        filename=f_path,
        model_name=None,
        is_image=True
    )
```

**Review by christinestraub — CHANGES_REQUESTED** [review]

File: unstructured_inference/inference/elements.py (hunk 6 above)
Comment (yuming-long) [reply unstructured_inference/inference/elements.py:216]: ah i see, thanks!

Comment (yuming-long) [reply unstructured_inference/inference/elements.py]: okay i updated this  with the remove of image parameter and looks like the output on individual blocks changed...
saw error on my unstructured branch: 
```
test_unstructured/partition/pdf-image/test_image.py::test_partition_image_hi_res_ocr_mode[individual_blocks] - AssertionError: assert 'LORE EAT O R...imeescAnelye’' == 'LayoutParser...mage Analysis'
  - LayoutParser: A Unified Toolkit for Deep Learning Based Document Image Analysis
  + LORE EAT O RES AA ERS: LER ee, A lodirininie: Hluset flocument imeescAnelye’
```
so the output is worse... i am sure why but am looking into it

yuming-long: okay this is not related to this branch, the change in unstructured is causing the error

Dockerfile
cragwolfe: yes, not sure this Dockerfile should remain in the first place

cragwolfe (approved):

PR #232, opened 2023-09-30
Feat/chipper v2

benjats07:
This PR adds chipper v2 to the library, leveraging inference speedup and support for hierarchy (assign a parent element to the elements)

The main library changes are
* New model, `chipperv2` is available. There are significant changes that necessitate code changes in pre and post-processing beyond just changes to the weights.
* a schema change to the `TextRegion` object, which now has a `bbox` property, rather than being a subclass of `Rectangle`
* `LayoutElement` now has a `parent` property that can be populated with another `LayoutElement` to represent inferred hierarchy.
* `LocationlessLayoutElement` is removed. Elements produced by Chipper now have bounding boxes, and if we need to handle elements without bounding boxes in the future, we can make the `bbox` property optional.

Note that because of the schema changes, ingest tests aren't working properly due to the incompatibility with the current `unstructured` version.

In order to test this new model you can use
```
from unstructured_inference.constants import OCRMode

file = "sample-docs/layout-parser-paper-fast.pdf"
model = get_model("chipperv2")
    file,
    model,
    ocr_mode=OCRMode.FULL_PAGE.value,
    ocr_strategy="never",
)

print(doc)
```

Also, you can try with an image:
```

file = "sample-docs/example_table.jpg"
model = get_model("chipperv2")
doc = layout.DocumentLayout.from_image_file(
    file,
    model,
    ocr_mode=OCRMode.FULL_PAGE.value,
    ocr_strategy="never",
)

print(doc)
```


cragwolfe:
@qued, a hugging face token cred needs to be set, right?

ajjimeno:
Either the token is set or the user needs to be logged in HF to have access to Chipper v2. We just tested it to be sure that it is the case.

0-hero:
Eagerly waiting for this release :)

unstructured_inference/inference/elements.py
```diff
@@ -216,6 +192,20 @@ def extract_text(
...
         return text
 
+    @classmethod
+    def from_coords(
+        cls,
+        x1: Union[int, float],
+        y1: Union[int, float],
+        x2: Union[int, float],
+        y2: Union[int, float],
+        **kwargs,
```
badGarnet, line 202: can we have the optional variables explicitly listed here please so that it can be helpful to devs using hints.

unstructured_inference/inference/layout.py
```diff
@@ -190,7 +189,7 @@ def __init__(
         self.number = number
         self.detection_model = detection_model
         self.element_extraction_model = element_extraction_model
-        self.elements: Collection[Union[LayoutElement, LocationlessLayoutElement]] = []
+        self.elements: Collection[LayoutElement] = []
```
badGarnet, line 192: maybe worth document the breaking change here?

unstructured_inference/models/chipper.py
```diff
@@ -1,173 +1,502 @@
...
+from transformers.generation.logits_process import LogitsProcessor
+
+from unstructured_inference.constants import Source
+from unstructured_inference.inference.elements import Rectangle
+from unstructured_inference.inference.layoutelement import LayoutElement
 from unstructured_inference.models.unstructuredmodel import UnstructuredElementExtractionModel
+from unstructured_inference.utils import LazyDict
 
-MODEL_TYPES = {
+MODEL_TYPES: Dict[Optional[str], Union[LazyDict, dict]] = {
```
badGarnet, line 19: should these be in constants?

unstructured_inference/models/chipper.py
```diff
@@ -1,173 +1,502 @@
...
-        self.processor: DonutProcessor = DonutProcessor(
-            image_processor=DonutImageProcessor(
-                do_resize=True,
-                size=(self.required_w, self.required_h),
-            ),
-            tokenizer=self.tokenizer,
+        if torch.cuda.is_available():
+            self.device = "cuda"
+        else:
+            self.device = "cpu"
```
badGarnet, line 58: auto detection is nice but I would suggest we also add into init the ability to specify device. And this logic is used to either provide default or change when specified device is not available

unstructured_inference/models/chipper.py
```diff
@@ -1,173 +1,502 @@
...
             ignore_mismatched_sizes=True,
             use_auth_token=auth_token,
         )
+        if swap_head:
+            lm_head_file = hf_hub_download(
+                repo_id=pre_trained_model_repo,
+                filename="lm_head.pth",
+                token=auth_token,
+            )
+            rank = 128
```
badGarnet, line 81: constant?

unstructured_inference/models/chipper.py
```diff
@@ -1,173 +1,502 @@
...
+        self.input_ids = self.processor.tokenizer(
+            prompt,
+            add_special_tokens=False,
+            return_tensors="pt",
+        ).input_ids.to(self.device)
+
+        self.start_tokens = [
+            v
+            for k, v in self.processor.tokenizer.get_added_vocab().items()
+            if k.startswith("<s_") and v not in self.input_ids
```
badGarnet, line 98: another constant?

unstructured_inference/models/chipper.py
```diff
@@ -1,173 +1,502 @@
...
+                # max_length=self.max_length,
+                do_sample=True,
+                top_p=0.92,
+                top_k=5,
+                no_repeat_ngram_size=0,
+                num_beams=3,
+                return_dict_in_generate=True,
+                output_attentions=True,
+                output_scores=True,
+                output_hidden_states=False,
```
badGarnet, line 139: remove the commented line? and maybe make a constant for this default config?

unstructured_inference/models/chipper.py
```diff
@@ -1,173 +1,502 @@
...
-            tokenizer=self.tokenizer,
+        if torch.cuda.is_available():
+            self.device = "cuda"
+        else:
+            self.device = "cpu"
+        self.max_length = max_length
+        self.heatmap_h = heatmap_h
+        self.heatmap_w = heatmap_w
+        self.source = source
+        self.processor = DonutProcessor.from_pretrained(pre_trained_model_repo, token=auth_token)
```
badGarnet, line 63: should we pass device down to Donut as well?

unstructured_inference/models/chipper.py
```diff
@@ -1,173 +1,502 @@
...
+        """
+        decoder_cross_attention: tuple(tuple(torch.FloatTensor))
+        Tuple (one element for each generated token) of tuples (one element for
+        each layer of the decoder) of  `torch.FloatTensor` of shape
+        `(batch_size, num_heads, generated_length, sequence_length)`
+        """
+        agg_heatmap = np.zeros([final_h, final_w], dtype=np.uint8)
+
+        for tidx in tkn_indexes:
+            hmaps = torch.stack(list(decoder_cross_attentions[tidx]), dim=0)
```
badGarnet, line 318: why do we need to cast into `list` first? Can we require this to just be higher dim tensor so that we don't need to do type casting?

unstructured_inference/models/chipper.py
```diff
@@ -1,173 +1,502 @@
...
+        heatmap_w: int = 30,
+        discard_ratio: float = 0.99,
+    ) -> List[int]:
+        """
+        decoder_cross_attention: tuple(tuple(torch.FloatTensor))
+        Tuple (one element for each generated token) of tuples (one element for
+        each layer of the decoder) of  `torch.FloatTensor` of shape
+        `(batch_size, num_heads, generated_length, sequence_length)`
+        """
+        agg_heatmap = np.zeros([final_h, final_w], dtype=np.uint8)
```
badGarnet, line 315: maybe use torch functions https://pytorch.org/docs/stable/generated/torch.zeros.html to ensure device consistency? (I __think__ np is on cpu always, right?)

badGarnet:
general comment:
- tested code working
- pr note instruction seem to be using an older version of inference interface? the following worked 
```

file = "sample-docs/layout-parser-paper-fast.pdf"
model = get_model("chipperv2")
doc = layout.DocumentLayout.from_file(file, model)
```
and confirmed also elements' source is `chipperv2`
- would really like to test and profile the code on gpu to profile device usage during inference but that can wait after this is in

unstructured_inference/inference/elements.py
qued: The purpose here is to support subclassing without having to override the classmethod whenever the constructor changes. If I add the optional kwargs from this class and keep the `**kwargs` for potential subclasses, would you feel good about that?

unstructured_inference/models/chipper.py
ajjimeno: OpenCV is being used, which unfortunately uses numpy. All the operations for bounding box detection are run on CPU, which should not be a speed issue.

unstructured_inference/models/chipper.py
ajjimeno: The max_length should be uncommented. It has been corrected.

unstructured_inference/models/chipper.py
ajjimeno: This would be nice not only for Chipper, since many GPUs might exist in the same system. This probably requires its own feature request.

ajjimeno:
I see many requests for constants. Could you provide an example of what is required in those requests? Is it a nice to have or a compulsory requirement? Thanks!

unstructured_inference/models/chipper.py
qued: This is consistent with how the initialization parameters are handled for other models (detectron2, yolox, etc.). It probably could be done better but maybe it could be a separate PR?

LaverdeS:
I haven't looked though the code but I threw some 15 pdf samples (mostly part of sample-docs and ingestion) using `chipperv2` to see if the hierarchical relationships were present and could not find any. 

The import `from unstructured_inference.constants import OCRMode` suggested in the [PR description](https://github.com/Unstructured-IO/unstructured-inference/pull/232#issue-1920392652) throws a `ImportError: cannot import name 'OCRMode' from 'unstructured_inference.constants'`, so the snippet I used is the one suggested by @ajjimeno in his [last comment](https://github.com/Unstructured-IO/unstructured-inference/pull/232#issuecomment-1751405613). 

I acknowledge the presence of `List` elements and the `parent` attribute of the `LayoutElements` but its value is always `None`, even when is certain that a hierarchy is present.

From the 15 pdf samples, 94 elements were detected in total but no `List-item` was present, which is also odd, considering `chipperv2` returns this category in the original XML output. I suspect this is related to the [`token2json`](https://github.com/Unstructured-IO/unstructured-inference/blob/b399869ddef55c4ebdc210fee6717cafb054327d/unstructured_inference/models/donut.py#L78C43-L78C43) transform from `DonutProcessor` @ajjimeno. This adds to the lack of `parent` attribute diff than `None`, a makes most of the returned documents to miss a considerable amount of the pdf content.

I tested more on a sample [russian.pdf](https://github.com/Unstructured-IO/unstructured-inference/files/12847938/russian.pdf).

Some observations:

This is the output of `chipperv2` in XML using it directly from HF:
image
This shows `chipperv2` successfully parses the document returning `List` and `List-item` elements in a hierarchical structure.

Now, this is the output of `chipperv2` being called through `unstructured-inference=='0.6.1'`:
image
All the content is present, but there is no chance for hierarchical info to fit in the element class data and the `List` elements are ignored.

With the last changes introduced in this PR, the output is:
image
Which is missing a lot of the content (all from which were `List`or `List-item`s in the XML), and has a suspicious `Image` element from `'source': <Source.PDFMINER: 'pdfminer'>`.

I am testing to plot to returned element coords but I have not clear what is the reference image for them. Are the coords returned considering the resolution needed for chipper when doing pdf2image? @ajjimeno.

unstructured_inference/inference/elements.py
qued: Made the above changes in [844868b](https://github.com/Unstructured-IO/unstructured-inference/pull/232/commits/844868b287f9b8360aaaad7da028a5d99028c057). LMK if this works.

unstructured_inference/inference/layout.py
qued: Added a comment in [16b3437](https://github.com/Unstructured-IO/unstructured-inference/pull/232/commits/16b343729ca1bdafbd373789314bd023b91eb897).

unstructured_inference/models/chipper.py
qued: Added a parameter that can be passed that overrides autodetection in [6aa76ab](https://github.com/Unstructured-IO/unstructured-inference/pull/232/commits/6aa76ab0536240849efd78f7b76ddbe06f750164). We'd have to think about how a user could actually pass this in, but the framework is there.

unstructured_inference/models/chipper.py
qued: Made the prefix a parameter in [a591613](https://github.com/Unstructured-IO/unstructured-inference/pull/232/commits/a5916134cdecfdc3452a478592efbc6ebfc67b91).

badGarnet:
> I see many requests for constants. Could you provide an example of what is required in those requests? Is it a nice to have or a compulsory requirement? Thanks!

So here is an example of using a settings constant to represent this parameters: https://github.com/Unstructured-IO/unstructured-inference/pull/232/commits/0871489c9d3279035d174f6e55b93334d72bbcb2 credit to @qued 
There are a few reasons:
- some values like `"<s_"` are used in multiple places so it is easier to maintain and less error prune if we define it once and use the variable name instead of typing the actual value
- some values may be a specific value that is really part of a configuration of the model like `rank=128`. Define it explicitly makes the code easier to read later on (so it is more clear on what knobs are present in the model and what values they are; otherwise someone would have to read through the code to find all those places and values)
- technically they are not required to make the model run but they would make maintenance much easier down the road so I guess you can call them nice to have. Hence I used question mark as a suggestion instead of requiring change on those.

unstructured_inference/inference/elements.py
badGarnet: maybe this is not the PR to discuss this fine print detail; this can be a follow up discussion on explicit vs. implicit patterns. The `kwarg` here make this class ambiguous: it is not clear from definition what parameters is can accept. If a subclass has need for more parameters for this function they obviously would just override this method so we need a newly defined function anyway. And we'd only pass those new parameters when using the subclass that is supposed to support. But this shouldn't block this PR I guess.

unstructured_inference/models/chipper.py
qued: I tried to implement this, but reverted it in [3ab9531](https://github.com/Unstructured-IO/unstructured-inference/pull/232/commits/3ab953144051135fbccb32dc079f3da595a17971). I don't think `DonutProcessor` uses torch models.

unstructured_inference/inference/elements.py
qued: Overriding the method isn't needed this way, that's the advantage of doing it this way (as evidenced by the fact that mostly it's the `LayoutElement` subclass of `TextRegion` that uses this method without overriding it)... but granted maybe it's not worth the ambiguity.

badGarnet (approved):

PR #242, opened 2023-10-06
Feat/download quantized model

benjats07:
After this PR is merged two main changes will be done:

* `onnxruntime` will be pinned to be <1.16 due some changes in quantization recently introduced not being compatible for the moment with the library and model.
*  yolox_quantized will be downloaded from HF instead being generated at runtime.

In order to test you can execute:
```

file = "sample-docs/example_table.jpg"
model = get_model("yolox_quantized")
doc = layout.DocumentLayout.from_image_file(
    file,
    model,
    ocr_strategy="never",
)
```

And should show a download from HF downloading a file called `yolox_l0.05_quantized.onnx`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,7 @@
+## 0.7.1
+
+* Download yolox_quantized from HF
+
 ## 0.7.0
 
 * Remove all OCR related code expect the table OCR code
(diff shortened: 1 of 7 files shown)


qued (approved):
LGTM!

PR #248, opened 2023-10-11
Bug when loading Chipper model

ajjimeno:


ajjimeno:
When running chipper, the public version, we could see the following error.

Traceback (most recent call last):
  File "/Users/ajimeno/Documents/git/unstructured-inference/test_chipper.py", line 4, in <module>
    model = get_model("chipper")
  File "/Users/ajimeno/Documents/git/unstructured-inference/unstructured_inference/models/base.py", line 63, in get_model
    model.initialize(**CHIPPER_MODEL_TYPES[model_name])
TypeError: initialize() missing 1 required positional argument: 'swap_head_hidden_layer_size'

qued (approved):

PR #249, opened 2023-10-11
build: improve packaging

qued:
PR to make to make packaging a little nicer, in line with how packaging works in `unstructured`. In particular, support for extras is added.

This consists of:
- Adding the file `sg.in` for dependencies related to `supergradients`. This file has been `pip-compile`d with output `sg.txt`.
- All dependencies have been `pip-compile`d.
- Adding the file `constraints.in` to add a `numpy` constraint that is required for compliance with `supergradients`. Dependencies that precede `sg.in` need to know about this to avoid conflicts. `constraints.in` is added as a constraints file for all other dependency files.
- `make` targets are added for installing `supergradients` dependencies.
- CI is updated to install `supergradients` dependencies.
- `supergradients` extra is added to `setup.py` and `MANIFEST.in` so it can be picked up by PyPi.
- To support the fact that the `supergradients` dependencies are not guaranteed to be installed, the way imports are handled in the `super_gradients.py` model file has been changed so imports are only attempted in the model constructor, in a module-level global variable. The constructor was used instead of doing the import in `initialize` due to an error that occurred during import when initiated from `initialize`.

#### Testing
From a fresh `pyenv` environment, and from the base folder of `unstructured-inference`:
```
pip install -e .
pip list | grep gradient
```
(There should be no output, as `supergradients` should not be installed.)

Now:
```
pip install -e ."[supergradients]"
pip list | grep gradient
```
(Output should show `super-gradients` is installed.)


pravin-unstructured (approved):
Looks good to me.

PR #250, opened 2023-10-11
chore: chipper model name should point to latest chipper version

qued:
Update model names so that `"chipper"` points to latest version of Chipper, while `"chipperv1"` points to old version.

### DIFF
--- a/unstructured_inference/constants.py
+++ b/unstructured_inference/constants.py
@@ -12,6 +12,7 @@ class Source(Enum):
     DETECTRON2_LP = "detectron2_lp"
     CHIPPER = "chipper"
     CHIPPERV1 = "chipperv1"
+    CHIPPERV2 = "chipperv2"
     PDFMINER = "pdfminer"
     MERGED = "merged"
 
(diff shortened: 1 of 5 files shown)


unstructured_inference/constants.py
```diff
@@ -11,7 +11,7 @@ class Source(Enum):
     DETECTRON2_ONNX = "detectron2_onnx"
     DETECTRON2_LP = "detectron2_lp"
     CHIPPER = "chipper"
-    CHIPPERV2 = "chipperv2"
+    CHIPPERV1 = "chipperv1"
```
cragwolfe, line 14: can we have 3 constants instead:
```
CHIPPER = "chipper"
CHIPPERV1 = "chippverv1"
CHIPPERV2 = "chipperv2"
```
to make this a bit more clear (and also more of a hint what will happen for v3)?

this has the implication of a dupe entry in MODEL_TYPES or making the translation of chipper -> latest version in https://github.com/Unstructured-IO/unstructured-inference/blob/da8b6e76e70193702c550e3027074169ea54a542/unstructured_inference/models/base.py#L62 , either of which is fine i think.

qued: Added another reference in [749c365](https://github.com/Unstructured-IO/unstructured-inference/pull/250/commits/749c365feba7fade6d753af69d0a60be0d5bb2bb).

cragwolfe (approved):

PR #251, opened 2023-10-12
Chore: remove layout parser dependency and detectron2

yuming-long:
### Summary

* Remove layout parser dependency since it will use paddle if paddle is installed, which can cause some error trace if it doesn't disable the trace handler
* Also remove `detectron2` and `UnstructuredDetectronModel` since they are wrapper for detectron from layout parser


yuming-long:
close this for now, not sure if we need to remove lp, can reopen if needed

PR #252, opened 2023-10-13
Faster version of Chipper

ajjimeno:


unstructured_inference/models/chipper.py
```diff
@@ -457,6 +484,47 @@ def __call__(self, input_ids: torch.LongTensor, scores: torch.FloatTensor) -> to
...
+        return any(
+            i
+            for la in _calc_banned_tokens(
+                input_ids,
+                num_batch_hypotheses,
+                self.repetition_window,
+                cur_len,
+            )
+            for i in la
+            if i not in self.skip_tokens
```
cragwolfe, line 524: this is not easy to read, why not just iterate through nested for loops the old-fashioned way?

ajjimeno: The loop looks like this now.

        for banned_tokens in _calc_banned_tokens(
            input_ids,
            num_batch_hypotheses,
            self.repetition_window,
            cur_len,
        ):
            for token in banned_tokens:
                if token not in self.skip_tokens:
                    return True

        return False

cragwolfe: thanks!

LaverdeS (approved):
LGMT! It is considerably faster now (X3) and it also fixed some `UncategorizedText` remnants/extras during generation.

cragwolfe:
Test instructions, how verified?

LaverdeS:
> Test instructions, how verified?

```
%%time

doc = DocumentLayout.from_file("layout-parser-paper-fast.pdf", detection_model=model)
for i in range(len(docb.pages)):
    print(f"********** Page {i}")
    for element in doc.pages[i].elements:
        display(element.type)
        display(element.text)
```

PR #253, opened 2023-10-13
Fix PDFMiner bug

ajjimeno:
Issue:

In some cases, PDFMiner identifies an image document as a full page and in other installations not. It is difficult to find out when PDFMiner behaves in one way or another. In either case tested, the version is `pdfminer.six v20221105`. The solution is to ignore any annotation coming from Chipper in case the full page clearing code is activated. Not sure if this is relevant to other models.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,7 @@
-## 0.7.4-dev0
+## 0.7.4-dev1
 
+* Fixed bug when PDFMiner predicts that an image text occupies the full page and removes annotations by Chipper.
+* Added random seed to Chipper text generation to avoid differences between calls to Chipper.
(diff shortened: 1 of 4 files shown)


LaverdeS:
LGTM. 

One remark: the output is complete but its very different between several calls of `chipper` for the same sample. This is expected from a generative model like `chipper`; nevertheless, this was not being that noticeable bofore, which makes me think that with the last changes introduced in this PR those differences could increase, meaning, chipper could be more inconsistent between generations and could have more hallucinations.

LaverdeS (approved):
LGTM

ajjimeno:
Added random seed to avoid the issue of differences across runs.

PR #256, opened 2023-10-16
Chore: allow table model to accept optional OCR data

yuming-long:
## Summary

Change `run_prediction` in the table model to accept optional OCR data for table OCR refactor as an alternative to getting OCR tokens in `get_tokens`.

## TODO

please see [CORE-2259](https://unstructured-ai.atlassian.net/browse/CORE-2259) to update `ocr_token` from dict to a data class.

[CORE-2259]: https://unstructured-ai.atlassian.net/browse/CORE-2259?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


unstructured_inference/models/tables.py
```diff
@@ -161,12 +161,17 @@ def run_prediction(
         self,
         x: Image,
         pad_for_structure_detection: int = inference_config.TABLE_IMAGE_BACKGROUND_PAD,
+        ocr_tokens: List = [],
     ):
         """Predict table structure"""
         outputs_structure = self.get_structure(x, pad_for_structure_detection)
-        tokens = self.get_tokens(x=x)
+        if ocr_tokens == []:
```
badGarnet, line 168: probably should allow it to be empty if the OCR detects no text. And we can set the default "not OCR'ed" state as `None`.
Additionally we may want to specify a bit more of the requirement for `ocr_tokens` is a list of dictionaries. For now we can just add the docstring to layout what is required for each token. And leave a note for future FIXME to refactor token data into a dataclass so we have clear expectations on what it should "look" like.

unstructured_inference/models/tables.py
```diff
@@ -161,12 +175,17 @@ def run_prediction(
...
         x: Image,
         pad_for_structure_detection: int = inference_config.TABLE_IMAGE_BACKGROUND_PAD,
+        ocr_tokens: Optional[List[Dict]] = None,
     ):
         """Predict table structure"""
         outputs_structure = self.get_structure(x, pad_for_structure_detection)
-        tokens = self.get_tokens(x=x)
+        if ocr_tokens is None:
+            logger.warning(
+                "Table OCR from get_tokens method will be deprecated.",
```
badGarnet, line 184:
```suggestion
                "Table OCR from get_tokens method will be deprecated.",
                "In the future the OCR tokens are expected to be passed in."
```

badGarnet (approved):
could you link the ticket associated with this PR in the PR note please?

PR #257, opened 2023-10-16
fix: invalid zoom lead to cv2 error

badGarnet:
- fix a bug where zoom factor <= 0 raises an exception at runtime
- now zoom factor <=0 will result in no scaling of the image

Related report from user: https://unstructuredw-kbe4326.slack.com/archives/C044N0YV08G/p1697456659162509

Test:
This PR adds a unit test with cases where zoom factor is 0 or -1. Without this fix the same error the user reported shows up:
```
error: (-215:Assertion failed) inv_scale_x > 0 in function 'resize'
```


benjats07 (approved):
LGTM

PR #258, opened 2023-10-17
fix: memory leak on chipper processor, beam search parameters, and bbox bug

ajjimeno:
This PR intends to solve the following issues:

* Memory leak in DonutProcessor when using large images in numpy format
* Use the right settings for beam search size > 1
* Solve a bug that in very rare cases made the last element predicted by Chipper to have a bbox = None

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,9 @@
+## 0.7.7
+
+• Fix a memory leak in DonutProcessor when using large images in numpy format
+• Set the right settings for beam search size > 1
+• Fix a bug that in very rare cases made the last element predicted by Chipper to have a bbox = None
+
 ## 0.7.6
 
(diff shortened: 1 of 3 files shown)


ajjimeno:
Example test snipper. Without the proposed changes it runs out of memory and if there is enough memory, it breaks because an element has no bbox. You need to use the proposed document from sample-docs, please change the location accordingly.

Probably on a Mac with a lot of memory the memory problem might not be a problem. I did not see the memory leak in a mac computer. On linux, with 16gb, it runs out of memory pretty quickly, e.g. on a c6i.2xlarge instance.

```

model = get_model("chipper")
doc = DocumentLayout.from_file("sample-docs/patent.pdf", model, pdf_image_dpi=300) 
```

cragwolfe (approved):

PR #260, opened 2023-10-17
Fix/no order chipper elements

benjats07:
This code can't be executed with the current version of `unstructured-inference` due a lack of coordinates on Picture elements produced by Chipper. 
However, those elements shouldn't be ordered at all, so the fix is to skip ordering when Chipper is used.

After this PR is merged this code is correctly executed.

```

file = "sample-docs/recalibrating-risk-report.pdf"
model = get_model("chipperv1")
    file,
    model,
)

print(doc)
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,7 @@
+## 0.7.10-dev0
+
+* fix: Skips ordering elements coming from Chipper
+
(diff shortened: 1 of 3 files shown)


ajjimeno (approved):
LGTM!

PR #261, opened 2023-10-18
fix: chipper memory problem with long documents on Intel CPUs

ajjimeno:


ajjimeno:
On an Intel machine with 16Gb of RAM and no swap, the following script will freeze the machine till the process is being killed. With the proposed changes, this would run with less memory.

```

model = get_model("chipper")
doc = DocumentLayout.from_file("sample-docs/layout-parser-paper.pdf", model, pdf_image_dpi=300) 
```

ajjimeno:
I think the CI / test_ingest errors are due to changes in the ingestion tests and not specifically due to the changes in this PR.

badGarnet:
confirming on intel instance this now uses reduce amount of memory (and actually can run a full pdf through where it wasn't able to due to OOM before)

badGarnet:
and just for record this is the memory profile:

badGarnet (approved):

PR #262, opened 2023-10-18
fix: annotate image is fixed after refactor

benjats07:
Currently the code for annotate() don't add details when add_details=True, for example:

```

file = "example-docs/layout-parser-paper-fast.pdf"
model = get_model("yolox_quantized")
doc = DocumentLayout.from_file(file,model)

for i,page in enumerate(doc.pages):
    page.annotate(add_details=True).save(f"{i}.png")
```

After this PR is merged, the resulting image should show origin for all elements using the same code.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
-## 0.7.10-dev0
+## 0.7.10-dev1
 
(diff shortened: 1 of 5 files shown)


ajjimeno (approved):
LGTM! I was not aware of this functionality.

PR #264, opened 2023-10-20
chore: streamline kwarg handling

qued:
Improves kwarg handling so that kwargs are handled explicitly when needed, and suppressed otherwise.

#### Testing:

CI should pass.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,4 @@
-## 0.7.10-dev2
+## 0.7.10
 
 * Handle kwargs explicitly when needed, suppress otherwise
 * fix: Reduce Chipper memory consumption on x86_64 cpus
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,6 @@
 ## 0.7.10-dev2
 
-* Handle kwargs related to pdf processing in signature
+* Handle kwargs explicitly when needed, suppress otherwise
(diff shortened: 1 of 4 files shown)


unstructured_inference/inference/layout.py
```diff
@@ -484,6 +470,8 @@ def process_file_with_model(
     fixed_layouts: Optional[List[Optional[List[TextRegion]]]] = None,
     extract_tables: bool = False,
     pdf_image_dpi: int = 200,
+    extract_images_in_pdf: bool = False,
+    image_output_dir_path: Optional[str] = None,
```
badGarnet, line 474: what's the guide line here? When do we explicit set and when do we implicitly trust a kwarg does something down the call stack? And why are those two explicit here? It doesn't seem those two parameters are modified in this function and they are just passing through?

CHANGELOG.md
```diff
@@ -1,5 +1,6 @@
-## 0.7.10-dev1
+## 0.7.10-dev2
```
cragwolfe, line 1: please make this is a release version

unstructured_inference/inference/layout.py
qued: These are explicit since we don't want them passed down to `get_model`, but we do want them passed down to `.from_file`

badGarnet (approved):

PR #267, opened 2023-10-24
Feat/chipper gpu float16

ajjimeno:
We added the automatic mixed precision for CPU, but this is needed as well for GPU running. This PR makes this possible for GPU. 

For testing, probably before and after the change on any long page.

```

model = get_model("chipper")
doc = DocumentLayout.from_file("sample-docs/layout-parser-paper.pdf",
                               detection_model=model, pdf_image_dpi=300)
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,7 @@
+## 0.7.11-dev0
+
+* fix: use automatic mixed precision on GPU for Chipper
+
 ## 0.7.10
 
 * Handle kwargs explicitly when needed, suppress otherwise
--- a/unstructured_inference/__version__.py
+++ b/unstructured_inference/__version__.py
@@ -1,1 +1,1 @@
-__version__ = "0.7.10"  # pragma: no cover
+__version__ = "0.7.11-dev0"  # pragma: no cover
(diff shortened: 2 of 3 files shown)


LaverdeS (approved):
LGTM. Tested on colab (login in to HF to use `chipperv2`) with a T4 using the snippet:
```
%%time
doc = %memit -i3 DocumentLayout.from_file("donut_rl_merged.pdf", detection_model="chipper", pdf_image_dpi=300)
```
Here are the results for `unstructured==0.7.10`:
```
peak memory: 5038.44 MiB, increment: 53.56 MiB
CPU times: user 16min 18s, sys: 34.1 s, total: 16min 52s
Wall time: 16min 36s
```
And the results for `unstructured==0.7.10-dev0` (with the changes in this PR):
```
peak memory: 4999.92 MiB, increment: 940.64 MiB
CPU times: user 17min 19s, sys: 36.8 s, total: 17min 56s
Wall time: 17min 37s
```
Both version executed a parsed a pdf document with 66 pages (academic paper) successfully. There is a big change in the memory `increment` value though, from 53.56 MiB to 940.64 MiB, but the GPU RAM looks better. In the image, the first part of the GPU RAM region until the red top ends, corresponds to `unstructured==0.7.10`, and the rest in orange is with `unstructured==0.7.10-dev0`.
image

PR #268, opened 2023-10-26
Feat/field to store inner elements

benjats07:
This PR introduces `clean_pdfminer_inner_elements` , which deletes pdfminer elements inside other detection origins such as YoloX or detectron.
This function returns the clean document and stores nested elements inside `pdfminer_inner_text`

The best way to check that this function is working properly is checking the new test test_clean_pdfminer_inner_elements in test_unstructured_inference/inference/test_processing_elements.py

Also, the new ingest-tests will reflect the deletion of the nested elements.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -109,6 +109,7 @@ jobs:
       uses: actions/checkout@v4
       with:
         repository: 'Unstructured-IO/unstructured'
+        ref: 'benjamin/feat/clean-by-store-pdfminer-inner-elements'
     - name: Checkout this repo
       uses: actions/checkout@v4
       with:
--- a/test_unstructured_inference/inference/test_processing_elements.py
+++ b/test_unstructured_inference/inference/test_processing_elements.py
@@ -1,0 +1,80 @@
+import pytest
+
+from unstructured_inference.constants import Source
+from unstructured_inference.inference.elements import Rectangle
+from unstructured_inference.inference.layout import DocumentLayout, LayoutElement, PageLayout
+
+# A set of elements with pdfminer elements inside tables
+deletable_elements_inside_table = [
+    LayoutElement(
+        bbox=Rectangle(0, 0, 100, 100),
+        text="Table with inner elements",
+        type="Table",
+    ),
+    LayoutElement(bbox=Rectangle(50, 50, 70, 70), text="text1", source=Source.PDFMINER),
+    LayoutElement(bbox=Rectangle(70, 70, 80, 80), text="text2", source=Source.PDFMINER),
+]
+
+# A set of elements without pdfminer elements inside
+# tables (no elements with source=Source.PDFMINER)
+no_deletable_elements_inside_table = [
+    LayoutElement(
+        bbox=Rectangle(0, 0, 100, 100),
+        text="Table with inner elements",
+        type="Table",
+        source=Source.YOLOX,
+    ),
+    LayoutElement(bbox=Rectangle(50, 50, 70, 70), text="text1", source=Source.YOLOX),
+    LayoutElement(bbox=Rectangle(70, 70, 80, 80), text="text2", source=Source.YOLOX),
+]
+# A set of elements with pdfminer elements inside tables and other
+# elements with source=Source.PDFMINER
+# Note: there is some elements with source=Source.PDFMINER are not inside tables
+mix_elements_inside_table = [
+    LayoutElement(
+        bbox=Rectangle(0, 0, 100, 100),
+        text="Table1 with inner elements",
+        type="Table",
+        source=Source.YOLOX,
+    ),
+    LayoutElement(bbox=Rectangle(50, 50, 70, 70), text="Inside table1"),
+    LayoutElement(bbox=Rectangle(70, 70, 80, 80), text="Inside table1", source=Source.PDFMINER),
+    LayoutElement(
+        bbox=Rectangle(150, 150, 170, 170),
+        text="Outside tables",
+        source=Source.PDFMINER,
+    ),
+    LayoutElement(
+        bbox=Rectangle(180, 180, 200, 200),
+        text="Outside tables",
+        source=Source.PDFMINER,
+    ),
+    LayoutElement(
+        bbox=Rectangle(0, 500, 100, 700),
+        text="Table2 with inner elements",
+        type="Table",
+        source=Source.YOLOX,
+    ),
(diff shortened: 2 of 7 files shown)


benjats07:
@christinestraub 
I addressed the issue you mentioned [here](https://github.com/Unstructured-IO/unstructured/pull/1808#issuecomment-1781745486)
I think the code belongs here, and the main library just use the function clean_pdfminer_inner_elements from here.

.github/workflows/ci.yml
```diff
@@ -109,6 +109,7 @@ jobs:
       uses: actions/checkout@v4
       with:
         repository: 'Unstructured-IO/unstructured'
+        ref: 'benjamin/feat/clean-by-store-pdfminer-inner-elements'
```
cragwolfe, line 112: just a reminder to remove this line before merging

test_unstructured_inference/inference/test_processing_elements.py
```diff
@@ -0,0 +1,80 @@
...
+from unstructured_inference.inference.elements import Rectangle
+from unstructured_inference.inference.layout import DocumentLayout, LayoutElement, PageLayout
+
+# A set of elements with pdfminer elements inside tables
+deletable_elements_inside_table = [
+    LayoutElement(
+        bbox=Rectangle(0, 0, 100, 100),
+        text="Table with inner elements",
+        type="Table",
+    ),
```
badGarnet, line 13: is this element left intentionally without a source? Do think that can be a valid test but want to make sure this is the intention.

unstructured_inference/inference/layout.py
```diff
@@ -153,6 +153,23 @@ def from_image_file(
             pages.append(page)
         return cls.from_pages(pages)
 
+    def clean_pdfminer_inner_elements(self):
+        """Clean pdfminer elements from inside tables and stores them in pdfminer_inner_text
+        property of the elements"""
+        for page in self.pages:
+            tables = [e for e in page.elements if e.type == "Table"]
+            for i, element in enumerate(page.elements):
+                if element.source == Source.PDFMINER:
```
badGarnet, line 162: lets do inverse condition then `continue` here to reduce indentation of the code; its a good practice to improve code readability in general. Especially here we have so many levels of indentation

unstructured_inference/inference/layout.py
```diff
@@ -153,6 +153,23 @@ def from_image_file(
...
+                    element_inside_table = [
+                        element.bbox.is_in(t.bbox, error_margin=15) for t in tables
```
badGarnet, line 164: is the margin something we want to make a kwarg into this function? How did we arrive with this number?

unstructured_inference/inference/layout.py
```diff
@@ -153,6 +153,23 @@ def from_image_file(
...
+                    ]
+                    if sum(element_inside_table) == 1:
```
badGarnet, line 166: so this catches when an element is inside and only inside one table? Could you please clarify this in the docstring?

unstructured_inference/inference/layout.py
```diff
@@ -153,6 +153,23 @@ def from_image_file(
...
+                        parent_table_index = element_inside_table.index(True)
+                        parent_table = tables[parent_table_index]
```
badGarnet, line 168:
```suggestion
                        parent_table = tables[element_inside_table.index(True)]
```

unstructured_inference/models/base.py
```diff
@@ -52,7 +52,7 @@ def get_model(model_name: Optional[str] = None, **kwargs) -> UnstructuredModel:
         initialize_params = {**DETECTRON2_ONNX_MODEL_TYPES[model_name], **kwargs}
     elif model_name in YOLOX_MODEL_TYPES:
         model = UnstructuredYoloXModel()
-        initialize_params = {**YOLOX_MODEL_TYPES[model_name], **kwargs}
```
badGarnet, line 55: why this change?

test_unstructured_inference/inference/test_processing_elements.py
```diff
@@ -0,0 +1,80 @@
...
+def test_clean_pdfminer_inner_elements(elements, lenght_extra_info, expected_document_lenght):
+    # create a sample document with pdfminer elements inside tables
+    page = PageLayout(number=1, image=None, layout=elements)
+    page.elements = elements
+    document_with_table = DocumentLayout(pages=[page])
+
+    # call the function to clean the pdfminer inner elements
+    document_with_table.clean_pdfminer_inner_elements()
+
+    assert len(document_with_table.pages[0].elements) == expected_document_lenght
```
badGarnet, line 80: since you got the text to denote where the box should be and what is outside can we do assert more clearly to check if the right pdfminer box is in the right table? And the right outside box is outside?

test_unstructured_inference/inference/test_processing_elements.py
```diff
@@ -0,0 +1,80 @@
...
+
+@pytest.mark.parametrize(
+    ("elements", "lenght_extra_info", "expected_document_lenght"),
+    [
+        (deletable_elements_inside_table, 1, 1),
+        (no_deletable_elements_inside_table, 0, 3),
+        (mix_elements_inside_table, 2, 5),
+    ],
+)
+def test_clean_pdfminer_inner_elements(elements, lenght_extra_info, expected_document_lenght):
```
badGarnet, line 71:
```suggestion
def test_clean_pdfminer_inner_elements(elements, length_extra_info, expected_document_length):
```

test_unstructured_inference/inference/test_processing_elements.py
```diff
@@ -0,0 +1,80 @@
...
+        type="Table",
+        source=Source.YOLOX,
+    ),
+    LayoutElement(bbox=Rectangle(0, 510, 50, 300), text="Inside table2", source=Source.PDFMINER),
+    LayoutElement(bbox=Rectangle(0, 550, 70, 400), text="Inside table2", source=Source.PDFMINER),
+]
+
+
+@pytest.mark.parametrize(
+    ("elements", "lenght_extra_info", "expected_document_lenght"),
```
badGarnet, line 64:
```suggestion
    ("elements", "length_extra_info", "expected_document_length"),
```

badGarnet (changes requested):

test_unstructured_inference/inference/test_processing_elements.py
benjats07: Yeah, was intentional, this is because if something is a table then should be kept

PR #276, opened 2023-11-03
enhancement: better error message on image/page extraction mismatch

qued:
Removed the reference to Poppler in error message when number of images extracted doesn't match the number of page layouts extracted (I'm no longer confident that this is caused by Poppler not being installed). Also provided some details in the message about how many of each were extracted, and updated the test associated with this error.

#### Testing:

Updated CI test should pass.


amanda103 (approved):
lgtm pending passing builds 🥳

PR #278, opened 2023-11-06
chore: remove logger info for chipper since its private

yuming-long:
Remove the logger info that user won't be able to see since chipper is only available in host API.


qued (approved):
Makes sense to me!

PR #282, opened 2023-11-06
Jj/warnings

### LINKED ISSUE #281: chore/ UserWarning raised whenever inference session started with Yolox model
The following warning is being raised any time an inference session is started with Yolox:
```
test_unstructured_inference/inference/test_layout.py::test_process_file_no_warnings
  /Users/<user>/.pyenv/versions/inference/lib/python3.10/site-packages/onnxruntime/capi/onnxruntime_inference_collection.py:54: UserWarning: Specified provider 'CUDAExecutionProvider' is not in available provider names.Available providers: 'CPUExecutionProvider'
    warnings.warn(
```

No warning should be raised unless there's an error.

Coniferish:
Closes #281 

### Summary
Introduces check to see what model providers are available locally before making an InferenceSession call. Providers are kept in order of preference.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,7 @@
+## 0.7.12-dev0
+
+* chore: supress UserWarning about specified model providers
+
 ## 0.7.11
 
 * chore: remove logger info for chipper since its private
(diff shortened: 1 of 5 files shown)


Coniferish:
@qued I'm not seeing “Ingest Test Fixtures Update PR” as a workflow option for fixing the fixtures on this PR. Can you run it or give me permissions?

That said, I'm not sure what should be changing in the fixtures from this PR. It looks like `/home/runner/work/unstructured-inference/unstructured-inference/test_unstructured_ingest/text-output/s3: IRS-form-1987.pdf.json.txt` is the culprit, but I'm not immediately seeing what's changing from the log

qued (approved):
LGTM! Verified that a call to an ONNX model no longer outputs the warning, and that the test fails on the old code.

PR #286, opened 2023-11-15
Feat: improve image extraction by supporting all types of image elements detected by detection models

### LINKED ISSUE #285: Enhancement: improve image extraction by supporting all types of image elements detected by detection models
As of now, when we extract images, we only extract elements with the type `Image`. The goal of this issue is to support extracting elements with types `Picture` and `Figure`.

christinestraub:
Closes #285.
### Summary
- support extracting elements with types `Picture` and `Figure`
- add a class `ElementType` for the element type constants and use the constants to replace element type strings

### Testing
PDF: [algebra-graph-level1-1.pdf](https://github.com/Unstructured-IO/unstructured-inference/files/13368976/algebra-graph-level1-1.pdf)

```

    filename="algebra-graph-level1-1.pdf",
    extract_images_in_pdf=True,
)
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,7 @@
-## 0.7.13-dev0
+## 0.7.13-dev1
 
+* refactor: add a class `ElementType` for the element type constants and use the constants to replace element type strings
+* enhancement: support extracting elements with types `Picture` and `Figure`
 * chore: supress UserWarning about specified model providers
 
 ## 0.7.12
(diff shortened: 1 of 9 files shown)


CHANGELOG.md
```diff
@@ -1,5 +1,7 @@
-## 0.7.13-dev1
+## 0.7.13-dev2
```
cragwolfe, line 1: let's make this a non-dev release so it can get integrated into unstructured sooner rather than later.

benjats07 (approved):
LGTM
In the future we may explore the possiblity of just renaming Picture and Figure to be Image and simplify this.

PR #287, opened 2023-11-15
Initial commit for removing super-gradients dependencies and allowing…

pravin-unstructured:
… general exported Onnx Object detections models to be used. Still figuring out how to build unit test.


qued (approved):
LGTM!

PR #290, opened 2023-11-16
Chore: nit of table init logger to show up the log info

yuming-long:
`"Loading the table structure model ..."` wasn't showing up in log since it was using `logging.info` not `logger.info`


qued (approved):
LGTM!

PR #292, opened 2023-11-20
Feat/improve chipper bounding boxes

ajjimeno:
Chipper bounding boxes largely exceed the area where the annotated text is. This PR intends solving this problem.

There are several fixes applied:

* Improved cross attention processing. Maps are filtered at the head level, which is normalised [0-1], this improves the bounding box definition.
* The correlation between the token index and cross attention map has been resolved. Before, with beam search size = 1 or 3  there were cases in which the cross attention map did not match the token being processed.

In addition, the empty areas of the bounding boxes have been cleaned and the overlaps between bounding boxes are resolved by identifying the largest margin that separates both bounding boxes, identified as the largest gap without text either in the horizontal or vertical directions.

Note: overlapping bounding boxes for child elements are not resolved for now. In the case of a list with list items, the list element will be affected by the overlapping resolution code.


LaverdeS:
4 cases occur when filling the collection, from which just the latest one creates completely valid and ready-to-use filled PDFs:

- CASE 1: no identified field_names (all = "f"): Always triggers `random-sentence-generator`. Valid but low quality.
- CASE 2: filled form including checkboxes, but no rendered content. Invalid (`todo:` confirm/fix).
- CASE 3: filled form and render but no checkboxes. Valid but not ready.
- CASE 4: filled form and render. Valid and ready.

LaverdeS:
I am getting some dependency errors for the use of `cv2.typing` and for the KW argument `token` when constructing the model with `transformers`: `PreTrainedModel.from_pretrained()`. The call:

`elements = partition(filename="sample-docs/RAND_RRA2977-1-8.pdf", strategy='hi_res', model_name="chipperv2")`

ajjimeno:
Thanks @LaverdeS I tried a new python environment and run make install and it worked. This is the code I tested.

```

model = get_model("chipper")


    "RAND_RRA2977-1.pdf",
    detection_model=model,
    pdf_image_dpi=300,
)
```

LaverdeS (approved):
The bboxes are considerably better with the changes introduced in this PR! nevertheless, some are still too off (I shared some samples internally) so I guess more incremental improvement is necessary. Considering this is not the end of these efforts, and that the changes are already showing improvement, I approve this PR counting on another being open in the near future.

ajjimeno:
Thanks @LaverdeS Currently, probably that is the best approximation using the attention mechanism. Additional work combining additional information might help.

PR #293, opened 2023-11-20
fix bug which allows label map from initialize params to be used from…

pravin-unstructured:
… reading in a json file.


qued (approved):
LGTM!

PR #295, opened 2023-11-23
Feat/chipper repetitions

ajjimeno:
In some cases Chipper repeats elements. This PR has additional mechanisms to detect these repetitions and provides mechanisms for filtering repetitions that cannot be identified during decoding.

Repetition detection:

* Tables benefit from beam search size = 3. When a table is detected using beam search size = 1, the generation restarts with beam search size = 3
* To avoid interacting with genuine repetitions, a context windows has been defined to avoid looking for repeated text in all the generated elements.
* Specific mechanism has been added to detect when repetitions happen in tables, which were not detected before.

Additional filtering:

* Remove empty tables
* Remove Picture elements that get repeated
* Remove repeated texts, this uses the bounding boxes and the matching on the elements text to identify repetitions

Repetitions are not easy to reproduce, but I would suggest selecting documents that are not part of Odetta annotation or available in this repository for additional testing. Additional testing in different environments (e.g. Linux, Mac OS) is recommended.


ajjimeno:
Here are two example images. One is processed as a table, since it was modelled like that in the ground truth, the problem is that the table is never finished, chipper generates unlimited "<td></td>" pair of tokens. The second has some repetitions that may happen due to the images in the document. With the proposed PR, these repetitions disappear.

The PDF document made Chipper generate repetitions but only under Linux.

Example code for the images. 

```

image_file_name = [change with image path]

model = get_model("chipper")
doc = DocumentLayout.from_image_file(
    image_file_name,
    detection_model=model,
)

print(*[element.__dict__ for element in doc.pages[0].elements])

print(*[element for element in doc.pages[0].elements], sep="\n")

```

Example code for the PDF file:

```

pdf_file_name = [change with PDF path]

model = get_model("chipper")
    pdf_file_name,
    detection_model=model,
    pdf_image_dpi=300,
 )

print(*[element.__dict__ for element in doc.pages[0].elements])

print(*[element for element in doc.pages[0].elements], sep="\n")

```


[RAND_RRA2977-1.pdf](https://github.com/Unstructured-IO/unstructured-inference/files/13482720/RAND_RRA2977-1.pdf)



42044487_0fb229dd-commoncrawl_blogannettepehrssonsezenit-b-volume-thirml_2

leah1985 (approved):

CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+## 0.7.21-dev
```
cragwolfe, line 1: probably should be a release again to get into unstructured

mengdih:
Tested with chipperv2, the repetition problem still exists. The issue could be from the new changes in main branch. Have discussed this with @ajjimeno .

ajjimeno:
Fixed the issue with the merging and added additional test cases. @mengdih please check again. Thanks!

mengdih (approved):

PR #301, opened 2023-12-11
fix: prevent image zoom too large for tesseract

badGarnet:
This PR addresses [CORE-2965](https://unstructured-ai.atlassian.net/browse/CORE-2965) by limiting zoom factor so that the scaled image can still be processed by tesseract.

## test

A unit test is added in this PR to test a unlikely case where we'd scale an image a few thousand times and massively exceed the limit without the fix.

Unstructured reviewers can also use the document in the ticket to test.

[CORE-2965]: https://unstructured-ai.atlassian.net/browse/CORE-2965?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/unstructured_inference/__version__.py
+++ b/unstructured_inference/__version__.py
@@ -1,1 +1,1 @@
-__version__ = "0.7.19-dev0"  # pragma: no cover
+__version__ = "0.7.19"  # pragma: no cover
--- a/unstructured_inference/__version__.py
+++ b/unstructured_inference/__version__.py
@@ -1,1 +1,1 @@
-__version__ = "0.7.18"  # pragma: no cover
+__version__ = "0.7.19-dev0"  # pragma: no cover
--- a/unstructured_inference/models/tables.py
+++ b/unstructured_inference/models/tables.py
@@ -16,6 +16,7 @@ from transformers import DetrImageProcessor, TableTransformerForObjectDetection
 
 from unstructured_inference.config import inference_config
 from unstructured_inference.constants import (
+    TESSERACT_MAX_SIZE,
     TESSERACT_TEXT_HEIGHT,
 )
 from unstructured_inference.inference.layoutelement import table_cells_to_dataframe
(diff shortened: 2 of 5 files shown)


unstructured_inference/__version__.py
```diff
@@ -1 +1 @@
-__version__ = "0.7.18"  # pragma: no cover
+__version__ = "0.7.19-dev0"  # pragma: no cover
```
cragwolfe, line 1: why not just release it?

unstructured_inference/models/tables.py
```diff
@@ -118,9 +119,13 @@ def get_tokens(self, x: Image):
                 text_height < inference_config.TESSERACT_MIN_TEXT_HEIGHT
                 or text_height > inference_config.TESSERACT_MAX_TEXT_HEIGHT
             ):
+                max_zoom = max(0, np.round(np.sqrt(TESSERACT_MAX_SIZE / np.prod(x.size) / 32), 1))
                 # rounding avoids unnecessary precision and potential numerical issues assocaited
                 # with numbers very close to 1 inside cv2 image processing
-                zoom = np.round(inference_config.TESSERACT_OPTIMUM_TEXT_HEIGHT / text_height, 1)
```
yuming-long, line 123: We also has duplicated code for this in [unst](https://github.com/Unstructured-IO/unstructured/blob/main/unstructured/partition/pdf_image/ocr.py#L464) and `get_tokens` method is not used in unst after OCR refactor. Are we ready to deprecate the code in inference? cc @cragwolfe @qued

cragwolfe: yes, we definitely want to move OCR code for tables to unstructured from unstructured-inference

yuming-long: @badGarnet lets move this to unst and i will deprecate all table OCR code in inference in separate PR :)

badGarnet: that's true; we can finish refactor first; and I will make a PR on unstructured repo

badGarnet:
moved here: https://github.com/Unstructured-IO/unstructured/pull/2252

PR #305, opened 2023-12-15
Refactor/remove analysis scripts

christinestraub:


scanny (approved):
LGTM, just one suggested change but approving in advance :)

PR #308, opened 2023-12-17
Feat/chipper v3

ajjimeno:
This new version of Chipper should largely improve the output for tables. In the attached file, the output looked as having many cells spread across multiple columns, and largely because of the $ character, which was inconsistently annotated in the Odetta set. As well colspan did not work properly for the header. This new version of Chipper does not predict thead and tbody tokens for tables.

To test it, you need to run the code below. It will print the predicted elements. The code should print only one page and one element. The element has a field name text_as_html. The HTML within that field can be pasted in a new file renamed as html to be open with a browser.

Example with Chipperv2

image

Example with Chipperv3

image

apple

```

model = get_model("chipper")
doc = DocumentLayout.from_image_file("[point to the location of the file]/apple.png", detection_model=model)

for i in range(len(doc.pages)):
          print(f"********** Page {i}")
          print(*[element.__dict__ for element in doc.pages[i].elements], sep="\n")
```


unstructured_inference/__version__.py
```diff
@@ -1 +1 @@
-__version__ = "0.7.19"  # pragma: no cover
+__version__ = "0.7.20-dev"  # pragma: no cover
```
cragwolfe, line 1: make at a real release and not a -dev one if planning to integrate with unstructured?

unstructured_inference/inference/layoutelement.py
```diff
@@ -108,7 +108,11 @@ def merge_inferred_layout_with_extracted_layout(
                 continue
         region_matched = False
         for inferred_region in inferred_layout:
-            if inferred_region.source in (Source.CHIPPER, Source.CHIPPERV1):
+            if inferred_region.source in (
+                Source.CHIPPER,
+                Source.CHIPPERV1,
```
qued, line 113: We might want to include `Source.CHIPPERV2` as well, if the intent here is "any chipper source". In fact we might want a constant that includes all chipper sources.

unstructured_inference/models/chipper.py
```diff
@@ -390,7 +401,11 @@ def deduplicate_detected_elements(
         min_text_size: int = 15,
     ) -> List[LayoutElement]:
         """For chipper, remove elements from other sources."""
-        return [el for el in elements if el.source in (Source.CHIPPER, Source.CHIPPERV1)]
+        return [
+            el
+            for el in elements
+            if el.source in (Source.CHIPPER, Source.CHIPPERV1, Source.CHIPPERV3)
```
qued, line 407: Same comment about also allowing `Source.CHIPPERV2`.

unstructured_inference/inference/layoutelement.py
```diff
@@ -237,7 +241,9 @@ def table_cells_to_dataframe(cells: dict, nrows: int = 1, ncols: int = 1, header
     return DataFrame(arr, columns=header)
 
 
-def partition_groups_from_regions(regions: Collection[TextRegion]) -> List[List[TextRegion]]:
+def partition_groups_from_regions(
+    regions: Collection[TextRegion],
+) -> List[List[TextRegion]]:
```
qued, line 246: This looks like it maybe shouldn't have been reformatted.

qued (commented):
Some comments.

unstructured_inference/inference/layoutelement.py
ajjimeno: Fixed, thanks!

unstructured_inference/inference/layoutelement.py
ajjimeno: Added a constant. Thanks!

unstructured_inference/models/chipper.py
ajjimeno: Added a constant. Thanks!

cragwolfe:
Please add test instructions, esp. how to test with the unstructured library. Ideally, the test instructions would include a doc where the output has improved.

unstructured_inference/models/chipper.py
```diff
@@ -46,6 +46,17 @@
...
+        "pre_trained_model_repo": "unstructuredio/chipper-v3",
+        "swap_head": True,
+        "swap_head_hidden_layer_size": 128,
+        "start_token_prefix": "<s_",
+        "prompt": "<s><s_hierarchical>",
+        "max_length": 1536,
+        "heatmap_h": 40,
+        "heatmap_w": 30,
+        "source": Source.CHIPPERV3,
+    },
```
cragwolfe, line 59: update the line 2 lines below so the default is "chipperv3"?

ajjimeno: Default mode is not Chipperv3

leah1985 (approved):
ran ttc & % missed words. % missed words are low at 0.069.

0-hero:
Is it possible to test this model?

PR #311, opened 2023-12-20
fix: typeerror when using chipper

### LINKED ISSUE #310: TypeError: '<' not supported between instances of 'NoneType' and 'float'
The API hit an error in the following line. `other` is set to `Rectangle(x1=0, y1=0, x2=None, y2=None)`, and the NoneTypes are blowing up the comparison. We need to:

* Protect against None in this function
* Determine why the coords are set like this and fix if needed

```
TypeError: '<' not supported between instances of 'NoneType' and 'float'
(34 additional frame(s) were not displayed)
...
  File "/home/notebook-user/unstructured/unstructured/partition/pdf_image/pdfminer_processing.py", line 65, in process_data_with_pdfminer
    merged_layout = merge_inferred_layout_with_extracted_layout(
  File "unstructured_inference/inference/layoutelement.py", line 119, in merge_inferred_layout_with_extracted_layout
    is_full_page_image = region_bounding_boxes_are_almost_the_same(
  File "unstructured_inference/inference/elements.py", line 289, in region_bounding_boxes_are_almost_the_same
    return region1.intersection_over_union(region2) > same_region_threshold
  File "unstructured_inference/inference/elements.py", line 110, in intersection_over_union
    intersection = self.intersection(other)
  File "unstructured_inference/inference/elements.py", line 94, in intersection
    x2 = min(self.x2, other.x2)
```

badGarnet:
This PR resolves #310 
- chipper, or any page layout extracted with an element extraction model do not have key attributes like `image_metadata` populated
- this leads to `None` values for image width and height, which lead to the bug
- this fix prevents the function early return after chipper finds the elements
- it continues the logic to allow other key attributes of the page to be filled
- a bonus from this fix is we remove the image data from the page (which is not needed downstream) for chipper generated pages

## Test

A unit test is modified to test all the routes, including using an element extraction model, for page layout

Additionally grab this attached pdf and when running partition using chipper the main branch would lead to type error but this fix would run without error.
[005-CISA-AA22-076-Strengthening-Cybersecurity-p1-p4.pdf](https://github.com/Unstructured-IO/unstructured-inference/files/13731533/005-CISA-AA22-076-Strengthening-Cybersecurity-p1-p4.pdf)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,7 @@
+## 0.7.21
+
+* fix: fix a bug where chipper, or any element extraction model based `PageLayout` object, lack `image_metadata` and other attributes that are required for downstream processing; this fix also reduces the memory overhead of using chipper model
+
 ## 0.7.20
 
 * chipper-v3: improved table prediction
--- a/test_unstructured_inference/inference/test_layout.py
+++ b/test_unstructured_inference/inference/test_layout.py
@@ -9,7 +9,10 @@ from PIL import Image
 
 import unstructured_inference.models.base as models
 from unstructured_inference.inference import elements, layout, layoutelement
-from unstructured_inference.inference.elements import EmbeddedTextRegion, ImageTextRegion
+from unstructured_inference.inference.elements import (
+    EmbeddedTextRegion,
+    ImageTextRegion,
+)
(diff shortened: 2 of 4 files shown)


ajjimeno (approved):
LGTM!

PR #314, opened 2024-01-02
Feat/chipper repetitions

ajjimeno:
In some cases Chipper repeats elements. This PR has additional mechanisms to detect these repetitions and provides mechanisms for filtering repetitions that cannot be identified during decoding.

Repetition detection:

- Tables benefit from beam search size = 3. When a table is detected using beam search size = 1, the generation restarts with beam search size = 3
- To avoid interacting with genuine repetitions, a context windows has been defined to avoid looking for repeated text in all the generated elements.
- Specific mechanism has been added to detect when repetitions happen in tables, which were not detected before.

Additional filtering:

- Remove empty tables
- Remove Picture elements that get repeated
- Remove repeated texts, this uses the bounding boxes and the matching on the elements text to identify repetitions

Here are two example images. One is processed as a table, since it was modelled like that in the ground truth, the problem is that the table is never finished, chipper generates unlimited "<td></td>" pair of tokens. The second has some repetitions that may happen due to the images in the document. With the proposed PR, these repetitions disappear.

The PDF document made Chipper generate repetitions but only under Linux.

Example code for the images.

```

image_file_name = [change with image path]

model = get_model("chipper")
doc = DocumentLayout.from_image_file(
    image_file_name,
    detection_model=model,
)

print(*[element.__dict__ for element in doc.pages[0].elements])

print(*[element for element in doc.pages[0].elements], sep="\n")
```

Example code for the PDF file:

```

pdf_file_name = [change with PDF path]

model = get_model("chipper")
    pdf_file_name,
    detection_model=model,
    pdf_image_dpi=300,
 )

print(*[element.__dict__ for element in doc.pages[0].elements])

print(*[element for element in doc.pages[0].elements], sep="\n")
```

[RAND_RRA2977-1.pdf](https://github.com/Unstructured-IO/unstructured-inference/files/13482720/RAND_RRA2977-1.pdf)

286089231-8fef08e1-b3c6-4795-8746-1256af5288e7

286091300-aff403ef-4488-4e8e-8029-c2afa43cb355

A previous PR had issues with unstructured when running some mini holistic documents. This was due to a problem with bounding boxes and as well with nested elements (e.g. List and List-items). An exception would happen when running the mini holistic documents below. The example code below allows testing it. You might need to have to use the main branch in unstructured since recently a bug related to layout elements was identified and a different exception might happen. The code prints the unstructured version, so it is possible to check which version is being used. The output of the processing is stored in the out.un.json file.

```
from unstructured import __version__
print(__version__.__version__)

import json
from unstructured.partition.auto import partition
from unstructured.staging.base import elements_to_json

elements = json.loads(elements_to_json(partition([FILENAME], strategy="hi_res", model_name="chipper")))

with open('out.un.json', 'w') as w:
    json.dump(elements, w)
```

[soldering-iron-manual.pdf](https://github.com/Unstructured-IO/unstructured-inference/files/13733445/soldering-iron-manual.pdf)

[TX-penal-code-T8-36-1-3.pdf](https://github.com/Unstructured-IO/unstructured-inference/files/13733451/TX-penal-code-T8-36-1-3.pdf)


unstructured_inference/models/chipper.py
```diff
@@ -168,8 +193,81 @@ def predict(self, image) -> List[LayoutElement]:
...
+                        elements_iou = iou(e1, e2)
+                        ratio = SM(None, e1.text, e2.text).ratio()
+                        if (
+                            e1.type != e2.type
+                            and elements_iou > 0.75
+                            and ratio > 0.99
+                            and e2 not in parents
+                        ):
+                            repeated_elements.append(e2)
+
```
badGarnet, line 235: 1. invert some logics to reduce nesting
2. compute length only once
3. put more checks before computing IOU to avoid unnecessary computation

```suggestion
        n_elements = len(elements)
        for i, e1 in enumerate(elements):
            if e1 in repeated_elements:
                continue
            for j in range(i + 1, n_elements):
                e2 = elements[j]
                if e1 == e2 or e2 in parents or e1.type == e2.type:
                    continue
                elements_iou = iou(e1, e2)
                ratio = SM(None, e1.text, e2.text).ratio()
                if elements_iou > 0.75 and ratio > 0.99:
                    repeated_elements.append(e2)

```

unstructured_inference/models/chipper.py
```diff
@@ -168,8 +193,81 @@ def predict(self, image) -> List[LayoutElement]:
...
+        elements = [element for element in elements if element not in repeated_elements]
+
         return elements
 
+    @staticmethod
+    def remove_repeated_picture_elements(elements):
+        """
+        remove empty repeated picture elements
+        """
+        picture_text_to_remove = [
```
badGarnet, line 245: so this list has Picture type element text that has been repeated more than 3 times. How does that translate to "empty repeated picture elements"? We could get a document with the same pictures or similar pictures with same text inside? Like a document showing examples of different fonts?

unstructured_inference/models/chipper.py
```diff
@@ -564,7 +663,7 @@ def element_overlap(
         ]
 
         for check_element in elements:
-            if check_element == element:
+            if check_element == element or not check_element.bbox:
```
badGarnet, line 666: we have encountered cases where the bbox is not `None` but some of the attributes are e.g., bbox is like `(x1=1, y1=0, x2=None, y2=None)`. This condition wouldn't validate it. Maybe we should add a method to validate bbox to both check it is not None and do some check like https://github.com/Unstructured-IO/unstructured-inference/blob/main/unstructured_inference/inference/elements.py#L86-L88

unstructured_inference/models/chipper.py
```diff
@@ -1012,19 +1121,69 @@ def __call__(
         """
         num_batch_hypotheses = scores.shape[0]
         cur_len = input_ids.shape[-1]
+        new_input_ids = input_ids[:, slice(-self.context_length, cur_len)]
+        new_cur_len = new_input_ids.shape[-1]
+
         return _no_repeat_ngram_logits(
-            input_ids,
-            cur_len,
+            input_ids[:, slice(-self.context_length, cur_len)],
```
badGarnet, line 1128: isn't this `new_input_ids`

unstructured_inference/models/chipper.py
badGarnet: actually we have a function for iou already: https://github.com/Unstructured-IO/unstructured-inference/blob/main/unstructured_inference/inference/elements.py#L113-L120

unstructured_inference/models/chipper.py
```diff
@@ -1068,8 +1230,27 @@ def __call__(
...
+        input_ids: torch.LongTensor,
+        scores: torch.FloatTensor,
+        **kwargs,
+    ) -> bool:
+        """
+        Check if the already generated tokens contain the target token_id
+        """
+        output = self.target_token_id in input_ids
+
+        return output  # self.target_token_id in input_ids
```
badGarnet, line 1249:
```suggestion
        return self.target_token_id in input_ids
```

unstructured_inference/models/chipper.py
```diff
@@ -1068,8 +1230,27 @@ def __call__(
...
+
+
 def _no_repeat_ngram_logits(
-    input_ids: torch.LongTensor,
+    input_ids: torch.Tensor,
```
badGarnet, line 1253: the input is just sliced but from the same tensor as before, right? So the datatype should stay as `LongTensor`?

badGarnet:
tested the PR with the soldering pdf locally and found compared to main (file 1) we got:
- repeated element for "Determining if Heating Element has Expired and Needs to be Replaced"
- repeated element for "C. A \"Beep\" means the element is good. No beep means the heating element has expired and needs replacing."
- revision of two figure caption elements (correctly) into Image elements
- better url text (no extra space)
[main.json](https://github.com/Unstructured-IO/unstructured-inference/files/14106366/main.json)
[pr.json](https://github.com/Unstructured-IO/unstructured-inference/files/14106367/pr.json)
[diff.txt](https://github.com/Unstructured-IO/unstructured-inference/files/14106370/diff.txt)

unstructured_inference/models/chipper.py
ajjimeno: Done, created the function and updated where it was relevant.

ajjimeno:
@badGarnet Just in case it is relevant. When comparing the main branch and this one, "chipper" points to "chipperv2" in the main branch and to "chipperv3" in this branch. Probably you realised about it, but just in case.

qued:
I confirmed what Yao found, that for the file [soldering-iron-manual.pdf](https://github.com/Unstructured-IO/unstructured-inference/files/13733445/soldering-iron-manual.pdf) on page 3, there are no repeated elements on the `main` branch while repeated elements are introduced on this branch. 

This is concerning because it's one example of the PR making the problem worse. It may be a one-off and this PR may fix more cases than it solves, but it suggests we need to look at the statistics to determine whether this makes things better or worse.

ajjimeno:
@qued @badGarnet repetition has been solved. There was a condition to move from the initial beam search size = 1 to beam search size = 3 that I tried to optimize. I reverted it to the original implementation of NGramRepetitonStoppingCriteria.

qued:
Closing, as I don't think this is actively being worked on. We can re-open if needed.

PR #318, opened 2024-01-25
feat: add support for partitioning .heic files

Coniferish:
.heic files are an image filetype we have not supported. This PR is the one of two for adding support. The other PR is in the `unstructured` repo: #[2454](https://github.com/Unstructured-IO/unstructured/pull/2454)


Coniferish:
Can someone help me resolve the failing test. This is the part that's failing in chipper.py:
```
            hmap = flat.view(heatmap_h, heatmap_w)

            hmap = hmap.unsqueeze(dim=-1).to(torch.float32).cpu().numpy()
            hmap = (hmap * 255.0).astype(np.uint8)  # type:ignore
            # (40, 30, 1) uint8
            # fuse heatmaps for different tokens by taking the max
            agg_heatmap = np.max(
                np.asarray(
                    [
                        agg_heatmap,
                        cv2.resize(
                            hmap,
                            (final_w, final_h),
                            interpolation=cv2.INTER_LINEAR_EXACT,  # cv2.INTER_CUBIC
                        ),
                    ],
                ),
                axis=0,
            ).astype(np.uint8)
```

`flat.view()` and `hmap.unsqueeze()` both return a `Tensor` object, which seems to be causing the issue. I'm not familiar with working with Tensor objects, though.

christinestraub (approved):
LGTM! 
What do you think about calling `register_heif_opener()` once right [after importing modules](https://github.com/Unstructured-IO/unstructured-inference/blob/main/unstructured_inference/inference/layout.py#L27), instead of calling it multiple times in `layout.py`?

christinestraub (changes requested):
I found that we don't need to install `pillow_heif` library in `unstructured-inference` repo if we call `register_heif_opener()` [before calling `unstructured-inference` functions](https://github.com/Unstructured-IO/unstructured/blob/main/unstructured/partition/pdf.py#L291) in unstructured repo.

PR #322, opened 2024-02-17
Fix: wrong label map key type

erjieyong:
When reading the json file via `UNSTRUCTURED_DEFAULT_MODEL_INITIALIZE_PARAMS_JSON_PATH` environment variable, the label_map keys have to be converted to `int` type before it can be referenced. Added code to always convert label_map keys to `int` type.


qued (approved):
LGTM. I see some formatting changes, which makes me think the formatting settings are different, but I think we can fix that in a follow-up PR.

Issue #303, opened 2023-12-13
bug/reopen temp file (pdf high_res)


FAbrahamDev:
**Describe the bug**
In line: https://github.com/Unstructured-IO/unstructured-inference/blob/d4785df3057717a023e3425ea8b8dcc49fa10c67/unstructured_inference/inference/layout.py#L362
is a tmp file created to pass as filename to `process_file_with_model` -> `DocumentLayout.from_file` -> `load_pdf` -> `extract_pages` (pdf_miner).
The extract_pages tries to read the file again with `open_filename(pdf_file, "rb") as fp:`.

Which results in a `PermissionError: [Errno 13] Permission denied: 'C:\\Users\\...\\AppData\\Local\\Temp\\tmpf9flca30'` under windows. 

Same error here: 
https://github.com/Unstructured-IO/unstructured/blob/d3a404cfb541dae8e16956f096bac99fc05c985b/unstructured/partition/pdf_image/ocr.py#L79

**To Reproduce**

```
import tempfile

# print operating system name
import os
print(os.name)


# Create a temporary file
with tempfile.NamedTemporaryFile() as tmp_file:
    # Write some data to the file
    tmp_file.write(b'Hello, world!')
    tmp_file.flush()  # Flush the buffer to make sure data is written

    # Get the name of the file
    file_name = tmp_file.name

    # Since the file is closed after the with block, we need to open it again for reading
    with open(file_name, 'r') as file:
        # Read the data from the file
        content = file.read()
        print("Content of the temp file:", content)
```

**Expected behavior**
I expect it not to crash :)

**Additional context**
Possible solution taken from here: https://stackoverflow.com/questions/39983886/python-writing-and-reading-from-a-temporary-file
```
def process_data_with_model(
    data: BinaryIO,
    model_name: Optional[str],
    **kwargs,
) -> DocumentLayout:
    """Processes pdf file in the form of a file handler (supporting a read method) into a
    DocumentLayout by using a model identified by model_name."""

    with tempfile.TemporaryDirectory() as td:
        f_name = os.path.join(td, "tmp_file")
        with open(f_name, "w") as tmp_file:
            tmp_file.write(data.read())
            tmp_file.flush()

        layout = process_file_with_model(
            f_name,
            model_name,
            **kwargs,
        )

    return layout
```

or another solution by gpt:
``` 
import tempfile

with tempfile.NamedTemporaryFile(delete=False) as tmp_file:
[...]

PR #323, opened 2024-02-19
fix read write tmpfile on windows

Blaxzter:
This is a low impact fix for reading still open temporary files on windows. 
https://github.com/Unstructured-IO/unstructured-inference/issues/303

### DIFF
--- a/unstructured_inference/inference/layout.py
+++ b/unstructured_inference/inference/layout.py
@@ -2,6 +2,7 @@ from __future__ import annotations
 
 import os
 import tempfile
+import random
 from pathlib import PurePath
 from typing import BinaryIO, Collection, List, Optional, Union, cast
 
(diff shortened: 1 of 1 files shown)


unstructured_inference/inference/layout.py
```diff
@@ -351,11 +352,14 @@ def process_data_with_model(
 ) -> DocumentLayout:
     """Processes pdf file in the form of a file handler (supporting a read method) into a
     DocumentLayout by using a model identified by model_name."""
-    with tempfile.NamedTemporaryFile() as tmp_file:
-        tmp_file.write(data.read())
-        tmp_file.flush()  # Make sure the file is written out
```
pawel-kmiecik, line 356: instead of creating a separate file, could you try with this smaller change:
```
with tempfile.NamedTemporaryFile() as tmp_file:
        tmp_file.write(data.read())
        tmp_file.flush() 
        tmp_file.seek(0)
         layout = process_file_with_model(...)
```

KristianMischke: > Nope, probably it's not this issue - I'd try to create tempfile with options:
> 
> ```python
> tempfile.NamedTemporaryFile(delete=True, delete_on_close=False)
> ```
> 
> According to windows recommendations from here: https://docs.python.org/3/library/tempfile.html

Note that this only exists for Python 3.12, the docs for even Python 3.11 do not have the `delete_on_close` parameter: https://docs.python.org/3.11/library/tempfile.html#tempfile.NamedTemporaryFile

The reason I'm commenting is because I'm also running into this issue on windows

FAbrahamDev: And wound it defeat the purpose of a temp file to leave it on device after things have been done with it?
I guess you could do something like this if you dont want to use a temp folder

```
def process_data_with_model(
    data: BinaryIO,
    model_name: Optional[str],
    **kwargs,
) -> DocumentLayout:
    """Processes pdf file in the form of a file handler (supporting a read method) into a
    DocumentLayout by using a model identified by model_name."""
    file_name = ''
    with tempfile.NamedTemporaryFile(delete=False) as tmp_file:
        tmp_file.write(data.read())
        tmp_file.flush()  # Make sure the file is written out
        file_name = tmp_file.name
        
    try:
            file_name,
            model_name,
            **kwargs,
        )
    finally:
        os.remove(file_name)

    return layout
```

Blaxzter:
Superseded by #376, which applies the same `TemporaryDirectory` fix on main, and issue #303 is now closed. Closing as obsolete.

PR #330, opened 2024-03-19
feat: Remove supergradients

qued:
We introduced super-gradients as a dependency some months ago, intending it to be a temporary measure. The intent is to utilize any super-gradients models by exporting as ONNX models.

This also has the short term benefit of increasing test coverage.

#### Testing:

CI should pass (except perhaps the test coverage check).


tabossert (approved):
LGTM

pravin-unstructured (approved):
Looks good to me. Any new mdoels should be onnx exports.

PR #333, opened 2024-04-03
Refactor/get model allow registering of new models with function calls

badGarnet:
This PR allows us to register new models and use them in the system:
- refactor of `get_model` so it relies on information stored in `model_class_map` and `model_config_map` to initialize a new model with a given model name (which is the key to both mappings)
- a new function `unstructured_inference.models.base.register_new_model` now allows adding new model definition to the class mapping and config mapping
- after calling register new model one can now call `get_model` with the new model name and get the new model

## testing

New unit tests should pass

### DIFF
--- a/unstructured_inference/models/base.py
+++ b/unstructured_inference/models/base.py
@@ -1,6 +1,8 @@
+from __future__ import annotations
+
 import json
 import os
-from typing import Dict, Optional, Tuple, Type, Union
+from typing import Dict, Optional, Tuple, Type
 
 from unstructured_inference.models.chipper import MODEL_TYPES as CHIPPER_MODEL_TYPES
 from unstructured_inference.models.chipper import UnstructuredChipperModel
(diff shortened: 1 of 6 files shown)


unstructured_inference/models/base.py
```diff
@@ -15,17 +15,41 @@
...
-    **{name: UnstructuredDetectronModel for name in DETECTRON2_MODEL_TYPES},
-    **{name: UnstructuredDetectronONNXModel for name in DETECTRON2_ONNX_MODEL_TYPES},
-    **{name: UnstructuredYoloXModel for name in YOLOX_MODEL_TYPES},
-    **{name: UnstructuredChipperModel for name in CHIPPER_MODEL_TYPES},
-}
+
+def get_default_model_mappings() -> (
+    Tuple[
+        Dict[str, Type[UnstructuredModel]],
+        Dict[str, Union[dict, LazyDict]],
```
scanny, line 28: If you add:
```python
from __future__ import annotations
```
at the top of the file you can use modern notation for union and tuple types like:
```
tuple[
    dict[str, type[UnstructuredModel]],
    dict[str, dict | LazyDict],
]
```
which does not require the imports from `typing` (for those types) and is consistent with how Python and IDEs display types in messages etc.

Btw, a bare `dict` is not a type. Here maybe it's `dict[str, Any]`.

scanny (approved):
LGTM. One suggestion on typing conventions.

unstructured_inference/models/base.py
badGarnet: ah; forgot about future... I was back and forth with lint there 😅

PR #334, opened 2024-04-10
feat: calculate layouts iou to filter out incorrect extracted layout

amadeusz-ds:


pawel-kmiecik:
@amadeusz-ds I think we can close this one?

amadeusz-ds:
Closed because didn't give good results

PR #335, opened 2024-04-15
Save table prediction in cells format

plutasnyy:
This pull request allows to return predictions in raw cell representation from table transformer. It will be later used in unstructured repository to save prediction in a cells format for simpler metrics calculation

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,7 @@
+## 0.7.28
+
+* feat: allow table transformer agent to return table prediction in not parsed format
+
 ## 0.7.27
 
 * fix: remove pin from `onnxruntime` dependency.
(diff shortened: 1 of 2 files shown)


badGarnet (commented):
- please add a test
- and bump version

badGarnet:
> For the unit tests:
> 
> * The 'cells' type was previously tested in this function https://github.com/Unstructured-IO/unstructured-inference/pull/335/files#diff-56b4d700e2ed72befcaf37f68f915270aacece7bfdba7ce12906bf8d6efcf485R926
> * I have added unit test for raising error when the type is wrong
> 
> Bumping version:
> 
> * Last released version is 0.7.25, the change-log was edited for 0.7.26. Is there any script to bump version you referring to? How can I bump version?
> 
> Not passing test_ingest:
> 
> * @badGarnet could you guide me how to debug this problem? I looked into logs and I do not see any stacktrace and errors

the version bump is just edit both the CHANGLOG.md and `unstructured_inference/__version__.py` to match. We could add a ci lint check to ensure those two are consistent.

The ingest test failing can be ignored for now (it is not required anyway); we should take a look and maybe remove this step as it has been out of date for a while.

badGarnet (approved):

PR #340, opened 2024-04-23
ci: remove ingest tests job from CI workflow

scanny:
**Summary**
These ingests tests have been failing for a while now and were not required for merge. As we have no immediate intention of fixing these and their value in this context is doubtful, remove them.


amadeusz-ds (approved):
LGTM

scanny:
@amadeusz-ds I believe Crag was inclined for us to try to fix these rather than remove them.

amadeusz-ds:
> @amadeusz-ds I believe Crag was inclined for us to try to fix these rather than remove them.

Anyone looked into this? I could work on that, but I don't see anything in the logs that might point to the source of the problem.

qued:
Since this PR, @MthwRobinson disabled the tests, so I'm going to close this as moot for now. (Perfectly fine to reopen if we want the tests completely removed.)

PR #347, opened 2024-05-20
feat: add global working dir config params

amadeusz-ds:


unstructured_inference/config.py
```diff
@@ -7,13 +7,26 @@
...
 
 import os
+import tempfile
 from dataclasses import dataclass
+from functools import lru_cache
+from pathlib import Path
+
+
+@lru_cache(maxsize=1)
+def _get_tempdir(dir: str) -> str:
```
pawel-kmiecik, line 17: one thing - `dir` is reserved as std function name. Just replace with `dirname` or smth.

PR #355, opened 2024-06-10
fix: table HTML generation corrected

pawel-kmiecik:
Currently unstructured-inference generates tables with missing markup:

```html
<table>
  <thead> <!-- missing <tr> after <thead> -->
    <th>header cell1</th>
    <th>header cell2</th>
  </thead>
  <tr> <!-- would be nice to add <tbody> if <thead> is used -->
    <td>body cell1</td>
    <td>body cell2</td>
  <tr>
<table>
```

When it should look like:

```html
<table>
  <thead>
    <tr>
      <th>header cell1</th>
      <th>header cell2</th>
    </tr>
  </thead>
  <tbody>
    <tr>
      <td>body cell1</td>
      <td>body cell2</td>
    <tr>
  </tbody>
<table>
```

Additionally, fixed `fill_cells` function which added redundant cells that break the HTML table layout when spanned cells were found.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,4 @@
-## 0.7.35-dev0
+## 0.7.35
 Fix syntax for generated HTML tables
 
 ## 0.7.34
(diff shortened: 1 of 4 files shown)


badGarnet (approved):
we can just make this 0.7.35 for release

PR #361, opened 2024-06-25
build(deps): Update requirements

qued:
It looks like there have been a number of upstream updates, so I want to get those pulled in.

### DIFF
--- a/requirements/base.in
+++ b/requirements/base.in
@@ -2,6 +2,7 @@
 layoutparser
 python-multipart
 huggingface-hub
+numpy<2
 opencv-python!=4.7.0.68
 onnx
 onnxruntime>=1.17.0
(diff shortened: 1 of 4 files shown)


christinestraub (changes requested):
We added a pin to `numpy` in [unstructured](https://github.com/Unstructured-IO/unstructured/blob/main/requirements/deps/constraints.txt#L57-L59) and [unstructured-api](https://github.com/Unstructured-IO/unstructured-api/blob/main/requirements/constraints.in#L8) because a breaking change in the latest version caused `make pip-compile` to fail. For more details, you can refer to https://unstructured-ai.atlassian.net/browse/OSS-31

christinestraub (approved):
LGLTM!

PR #365, opened 2024-07-08
Dissable tatr loading warning

plutasnyy:
Relates to:
https://github.com/Unstructured-IO/unstructured/issues/3288

How to test:
Run
```
from unstructured_inference.models.tables import load_agent
load_agent()
```
There should be no warning about some weights not being loaded

### DIFF
--- a/unstructured_inference/models/tables.py
+++ b/unstructured_inference/models/tables.py
@@ -65,11 +65,10 @@ class UnstructuredTableTransformerModel(UnstructuredModel):
 
         try:
             logger.info("Loading the table structure model ...")
-
+            cached_current_verbosity = logging.get_verbosity()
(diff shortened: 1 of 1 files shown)


unstructured_inference/models/tables.py
```diff
@@ -65,7 +65,11 @@ def initialize(
 
         try:
             logger.info("Loading the table structure model ...")
+
+            logging.set_verbosity_error()
```
pawel-kmiecik, line 69: it would be nicer to:
- cache the current verbosity in a variable with [get_verbosity](https://huggingface.co/docs/transformers/en/main_classes/logging#transformers.utils.logging.get_verbosity)
- set the verbosity to the cached value with `set_verbosity`

PR #366, opened 2024-07-09
[Windows] Support Windows friendly tempfile access

tjtanaa:
# Description
Windows does not support multiple file access on the same file. Thus after creating the file, Windows will expect the content to be flushed into the file and exit the file object context. Only then the subsequent operation can open the file.

This addressed the issue https://github.com/Unstructured-IO/unstructured-inference/issues/303


Bill-XU:
hello, is there anybody in charge of this review?
this one is pretty urgent, please kindly give your help~~

cragwolfe:
CC @scanny

scanny:
This is fixed by #376.

Thanks for reporting @tjtanaa :)

PR #367, opened 2024-07-09
Feature/pdfinfo win

Falven:
On windows, poppler's pdfinfo utility requires the file to have an extension. This PR adds the extension to the temporary file such that pdfinfo is able to read the file.


Falven:
Could someone please review and approve? Without this fix unstructured will not work on windows.

PR #376, opened 2024-08-01
fix(layout): Windows support

### LINKED ISSUE #303: bug/reopen temp file (pdf high_res)

scanny:
**Summary**
Remove use of `tempfile.NamedTemporaryFile` which attempts to access the temporary file by name. This fails on Windows. Use `tempfile.TemporaryDirectory` instead.

Fixes #303

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,7 +1,8 @@
-## 0.7.37-dev1
+## 0.7.37-dev2
 
 * refactor: remove layout analysis related code
 * enhancement: Hide warning about table transformer weights not being loaded
+* fix(layout): Use TemporaryDirectory instead of NamedTemporaryFile for Windows support
 
 ## 0.7.36
 
--- a/unstructured_inference/__version__.py
+++ b/unstructured_inference/__version__.py
@@ -1,1 +1,1 @@
-__version__ = "0.7.37-dev1"  # pragma: no cover
+__version__ = "0.7.37-dev2"  # pragma: no cover
--- a/unstructured_inference/inference/layout.py
+++ b/unstructured_inference/inference/layout.py
@@ -3,7 +3,7 @@ from __future__ import annotations
 import os
 import tempfile
 from pathlib import PurePath
-from typing import BinaryIO, Collection, List, Optional, Union, cast
+from typing import Any, BinaryIO, Collection, List, Optional, Union, cast
 
 import numpy as np
 import pdf2image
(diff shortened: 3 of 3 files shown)


Coniferish (approved):
LGTM. 
Do we need to add any tests for this?

scanny:
It doesn't change the behavior, so I'd say no.

PR #386, opened 2024-09-12
add min version pin for protobuf to match unstructured repo

Coniferish:
This PR stemmed from removing the `protobuf` pin in `unstructured`. It turned out to be unnecessary to set a pin in this repo, so this PR simply updates the dependencies and addresses some errors that arose during CI (updating the `ruff check` command used in the Makefile and some type hints that were blocking the linting check in CI).


MthwRobinson (approved):
LGTM!

PR #390, opened 2024-10-10
use field to set default empty array

badGarnet:
Fix an issue where numpy array is used as default value to dataclass attribute. Similar to https://github.com/Unstructured-IO/unstructured-inference/pull/388 but for the LayoutElements class


ds-filipknefel (approved):
LGTM

PR #393, opened 2024-10-15
fix: fix bugs in layoutelements

badGarnet:
This PR fixes two bugs:

- fix a type casting issue when subtracting a int array with a float. This popped up when testing with `unstructured`, and some sources of element coordinates are of `int` type. This PR adds a new unit test case for `int` coord type with the grouping function
- fix element class id 0 becomes None bug: this happens when dumping `LayoutElements` as a list of `LayoutElement`. When an element class id is 0 the logic on main would treat it as no existing and use `None` as the type.


christinestraub (approved):
LGTM!

PR #396, opened 2024-10-18
fix missing source

badGarnet:
This PR fixes a bug that was found when working with `unstructured`: https://github.com/Unstructured-IO/unstructured/actions/runs/11403980075/job/31732726752#step:6:1778

## test

- this PR should fix the failing test above in `unstructured` ci
- this PR expands the test on `clean_layoutelements` to test `source` is kept

## note for release

Since we have not tagged a release for 0.7.42 this PR opt to not increase version number and include itself as part of 0.7.42


unstructured_inference/inference/layoutelement.py
```diff
@@ -88,7 +90,7 @@ def concatenate(cls, groups: Iterable[LayoutElements]) -> LayoutElements:
             element_probs=np.concatenate(probs),
             element_class_ids=np.concatenate(class_ids),
             element_class_id_map=class_id_map,
-            source=group.source,
+            source=sources[0],
```
christinestraub, line 93: It will raise an error if `sources` is an empty list (which means that `group.source` is not set for any `group` in `groups`)

PR #401, opened 2025-01-08
fix: update ONNX provider search and warn if GPU will not be used

dhdaines:
`yolox.py` was using a rather suspicious way of getting the available ONNX runtime providers, when there is a perfectly good function called `onnxruntime.get_available_providers()` that it could be using instead.

Also make it warn the user if CUDA is not present since they need to explicitly install a particular version of `onnxruntime-gpu` for this to work (see https://github.com/Unstructured-IO/unstructured/issues/2506 for details)

### DIFF
--- a/unstructured_inference/models/yolox.py
+++ b/unstructured_inference/models/yolox.py
@@ -4,9 +4,9 @@
 # https://github.com/Megvii-BaseDetection/YOLOX/blob/ac379df3c97d1835ebd319afad0c031c36d03f36/yolox/utils/demo_utils.py
 
 import cv2
+import logging
 import numpy as np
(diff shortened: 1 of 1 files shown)


cragwolfe:
@dhdaines , please update to use https://github.com/Unstructured-IO/unstructured-inference/blob/main/unstructured_inference/logger.py . thanks for the contribution!

dhdaines:
> @dhdaines , please update to use https://github.com/Unstructured-IO/unstructured-inference/blob/main/unstructured_inference/logger.py . thanks for the contribution!

Done!  Sorry it took me a while to get around to it...

PR #402, opened 2025-01-14
fix: fix bugs in data structure

badGarnet:
- fix bug when an empty list is passed into `TextRegions.from_list`
- fix bug when concatenating a list of `LayoutElements` the class id maps is not updated correctly


MaksOpp (approved):
LGTM

PR #403, opened 2025-01-17
remove layoutparser lib

tbs17:
this PR is to remove the layoutparser lib as we no longer rely on it anymore

hence the README.md is also updated to drop the note on supporting layoutparser model zoo


ajjimeno (approved):
LGTM!

PR #392, opened 2024-10-15
Add password with PDF files

pprados:
Add password with PDF files.
Must be combined with [PR 3721 in unstructured](https://github.com/Unstructured-IO/unstructured/pull/3721)

PR #405, opened 2025-01-19
Add password

Coniferish:
See https://github.com/Unstructured-IO/unstructured-inference/pull/392


unstructured_inference/inference/layout.py
```diff
@@ -51,6 +51,7 @@ def from_file(
         filename: str,
         fixed_layouts: Optional[List[Optional[List[TextRegion]]]] = None,
         pdf_image_dpi: int = 200,
+        password: Optional[str] = None,
```
badGarnet, line 54: the best practice is to avoid passing password as a plain text object since this leaves the password open for grabs. I'd suggest at least try hashing/encryption

Coniferish: @pprados

PR #407, opened 2025-01-22
feat: add back source attribute for backward compatibility

badGarnet:
This PR adds `source` back to `TextRegions` and `LayoutElements` for backward compatibility.


tbs17 (approved):
LGTM!

PR #415, opened 2025-03-05
unpin pdfminer-six

badGarnet:
- higher versions can have bugs that user have reported
- unpin it in the base.in so leaves other packages and user's constrains to resolve it


Coniferish (approved):
Approving in advance of checks passing, but LGTM

PR #417, opened 2025-04-03
Fix pdfminer-six dependencies.

pprados:
Pdfminer has a bug that prevents some PDF files from being read. I've fixed it. A new version is available and should be used in the various projects that depend on this component.

See https://github.com/pdfminer/pdfminer.six/pull/1081

This PR use the version >=20250327

To pass the CI/CD test, I also set the dependencies for python 3.9.

With a minimum version of 3.10, it will be possible to download the latest versions of dependencies.


pprados:
@badGarnet

The "Configure AWS credentials" failed => Input required and not supplied: aws-region

`make pip-compile` invocation fails. Dependencies are not correct for python 3.9. Raising the minimum version to 3.10 would solve the problem. Many dependencies now need 3.10.

The `pip-compile` approach doesn't handle this properly. An approach using Poetry, for example, would be preferable.

pprados:
@badGarnet
Can you fix the CI/CD ?

badGarnet:
> @badGarnet Can you fix the CI/CD ?

could you update the entry to `aws-region` to `us-east-2` in your PR?

badGarnet:
> @badGarnet Another bug in CI/CD: #420

@pprados this seem to be a result of updated dependencies. I will take a closer look at the PR and see what's changed. For reference [this](https://github.com/Unstructured-IO/unstructured-inference/actions/runs/14476550800/job/40603752955) was recently ran successfully.

pprados:
Cool. Thanks.

badGarnet:
ok so a few thoughts here:
- we avoid pinning with `==` as much as possible in `.in` files so that the library has as much compatibility with other libraries as possible; in fact it is due to other users' request that we relaxed `pdfminer` to be unconstrained
- if a project needs `pdfminer==v20250327` and `numpy>=2.0` one can simply add that constrain to the project's requirements and pip or uv or any dep management system can compile that into workable list of dependencies. e.g., write something like `requirements.in` for a project with 
```
unstructured-inference
pdfminer==v20250327
numpy>=2.0
```
then compile it with `pip compile requirements.in` will get `requirements.txt` that has the right version of `pdfminer` as well as the rest of dependencies.
- managing deps based on python version is not something `pip-tool` supports. Other popular tools like `uv` also doesn't support such usage so from maintenance point of view it is risky to attempt managing multiple versions of requirements based on python versions and risking confusing devs. 

Based on those I would suggest take option 2 to see if it fits your need and if it does we can close this PR.

pprados:
ok for me.

pprados:
It's ok for me.

PR #422, opened 2025-05-13
Downgrading transformers

ajjimeno:
Downgrading transformers to 4.40.0 to try preventing `Tensor on device meta is not on the expected device cpu!` issue.


ajjimeno:
Probably only the update in core-product was needed. Closing this PR since there are many updates needed to support python 3.9.

PR #430, opened 2025-06-02
feat: use thread lock to prevent racing condition

badGarnet:
This PR puts table model initialization behind a threadlock so it is thread safe. 

## testing
Use the following script to test (after changing extension) loading model in multithreading env.

[test_threading.txt](https://github.com/user-attachments/files/20560179/test_threading.txt)


ajjimeno (approved):
LGTM! I tested it with several python versions.

PR #435, opened 2025-06-30
fix: Do not hardcode file extension on temp files

awalker4:
This is a minor fix to improve our logging. When we buffer a file like input to disk in `process_data_with_model`, we always use the name `document.pdf`. This confused me when I found this in our logs:

```
2025-06-30 17:02:01,906 unstructured_inference INFO Reading image file: /var/folders/5k/frv076q97yl0ywybmzydhbsr0000gn/T/tmpc0uq7zde/document.pdf ...
2025-06-30 17:02:01,951 unstructured_api ERROR cannot identify image file '/private/var/folders/5k/frv076q97yl0ywybmzydhbsr0000gn/T/tmpc0uq7zde/document.pdf'
```

This path can be either pdfs or images, so let's just drop the extension to save ourselves some confusion.

Also added a comment so we don't forget why it's using a temp dir, not a temp file.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,7 @@
+## 1.0.7
+
+* Fix a hardcoded file extension causing confusion in the logs
+
 ## 1.0.6
 
 * Add slicing through indexing for vectorized elements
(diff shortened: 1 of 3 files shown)


qued (approved):
LGTM!

PR #443, opened 2025-10-07
enhancement: optimize outputs to objects metbo2xp

qued:
From codeflash: Optimizes table model pipeline by keeping the data as a tensor for longer, enabling vector operations and reducing conversions.


qued:
@claude critique this PR please.

PR #508, opened 2026-04-20
fix: fix rare occasion of initialization error by converting LazyDict into dict

badGarnet:
## Summary

  Fix `"argument after ** must be a mapping, not LazyDict"` raised from `get_model()` in some production environments.

  `get_model()` was doing `model.initialize(**initialize_params)` where `initialize_params` is a `LazyDict`. `**` unpacking calls `.keys()` on
  the mapping, which `LazyDict` inherits from `collections.abc.Mapping` — but we've seen environments where that inherited method isn't resolved
   at call time, surfacing as the error above.

  The fix materializes `initialize_params` into a plain `dict` via `__iter__` + `__getitem__` (both defined directly on `LazyDict`) before the
  `**` unpack. `LazyEvaluateInfo` values are still resolved lazily via `__getitem__`, so download-on-first-use semantics are preserved.

  ## Changes

  - `unstructured_inference/models/base.py` — normalize `initialize_params` to a plain dict before `model.initialize(**...)`.
  - `test_unstructured_inference/models/test_model.py` — new regression test `test_get_model_with_lazydict_config` exercising the `LazyDict`
  path with a `LazyEvaluateInfo` value.
  - `CHANGELOG.md` + `__version__.py` — bump to `1.6.7`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,12 @@
+## 1.6.7
+
+### Fix
+- `get_model` now materializes `LazyDict` model configs into a plain dict before
+  unpacking into `initialize(**...)`. Uses `__iter__` + `__getitem__` to avoid
+  depending on `Mapping.keys()`, which has been observed to fail at `**`
+  unpacking with "argument after ** must be a mapping, not LazyDict" in some
+  deployment environments.
+
 ## 1.6.6
 
 ### Enhancement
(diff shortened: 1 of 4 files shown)


tabossert (approved):
LGTM

vladimir-kivi-ds (approved):

PR #514, opened 2026-06-09
fix: render rotated PDF pages consistently with extracted text

badGarnet:
## Problem

For PDF pages with a non-zero `/Rotate`, the rendered page image and pdfminer's extracted text ended up in different coordinate frames. `convert_pdf_to_image` re-applied the page `/Rotate` with PIL on top of pypdfium2's render — but pypdfium2 already renders the display frame, so the extra rotation effectively returned the image to the un-rotated MediaBox frame. pdfminer, meanwhile, honors `/Rotate` and reports coordinates in the display frame. The two were then off by the page rotation, which scrambled the merged layout downstream.

## Fix

- `convert_pdf_to_image` no longer re-applies `/Rotate` on top of pypdfium2's render. The rendered image now shares pdfminer's display frame by default.
- For `/Rotate != 0` pages, the dominant text orientation is estimated from pdfminer character matrices (the only signal that reliably matches the rendered image). When a single 90-degree orientation clearly dominates — controlled by the new `PDF_ROTATION_DOMINANT_ANGLE_THRESHOLD` (default `0.95`) — the rendered image is rotated so its text is upright.
- The applied correction is exposed as `pdf_rotation_correction` in the page image metadata (PIL `.info` and PNG text), so downstream consumers can rotate extracted coordinates by the same angle and stay aligned.
- Orientation estimation failures degrade gracefully to no correction (image stays in its native display frame — still consistent with pdfminer, just possibly not upright).

### Guards
1. Only pages with a non-zero `/Rotate` are considered for correction.
2. A correction is applied only when one 90-degree orientation accounts for at least the threshold share of characters; ambiguous pages are left untouched.

The correction is applied to the image and reported for the coordinates together, so it can only affect uprightness, never the consistency between the layers.

## Tests
- Existing rotation test still passes (now via the correct render-then-correct mechanism).
- Added coverage for the no-correction path on unrotated pages and for the threshold guard skipping correction below the configured share.

Paired with the corresponding change in `unstructured` that mirrors `pdf_rotation_correction` onto the extracted coordinates.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

---
## Summary by cubic
Fix rotated PDF rendering by removing duplicate rotation and applying a text-based correction. Images and extracted coordinates now align with `pdfminer`, pages become upright when a clear orientation is detected, and `pdfminer.six` is now declared so the correction always runs.

- **Bug Fixes**
  - Stop re-applying page `/Rotate` in `convert_pdf_to_image`; `pypdfium2` already renders the display frame matching `pdfminer` coordinates.
  - For `/Rotate != 0` pages, estimate the dominant 90° text direction from `pdfminer` characters and rotate only if it meets `PDF_ROTATION_DOMINANT_ANGLE_THRESHOLD` (default 0.95); ambiguous pages are left as-is.
  - Store the applied angle as `pdf_rotation_correction` in PNG/PIL info and surface it in layout metadata so downstream consumers can rotate coordinates to match.

- **Dependencies**
  - Add `pdfminer.six` (>=20231228) and move imports to module level to fail fast if missing; refresh lockfile.

<sup>Written for commit b6c6c46c5b699810257cec02e1dd5fe29672ca82. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/Unstructured-IO/unstructured-inference/pull/514?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,8 @@
+## 1.6.12
+
+### Fixes
+- **Render rotated PDF pages consistently with extracted text**: `convert_pdf_to_image` no longer re-applies the page `/Rotate` on top of pypdfium2's render (which already produces the display frame), so the rendered image stays in the same coordinate frame as pdfminer's extracted text. For pages with a non-zero `/Rotate`, the dominant text orientation is detected and, when a single 90-degree orientation clearly dominates (configurable via `PDF_ROTATION_DOMINANT_ANGLE_THRESHOLD`), the image is rotated so its text is upright. The applied correction is exposed as `pdf_rotation_correction` in the page image metadata so downstream consumers can keep extracted coordinates aligned.
+
 ## 1.6.11
 
 ### Enhancement
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -37,6 +37,8 @@ dependencies = [
     "pandas>=1.5.0",
     "scipy>=1.17.0",
     "pypdfium2>=5.0.0",
+    # Used to read per-character matrices for PDF page text-orientation correction
+    "pdfminer.six>=20231228",
 ]
 
 [project.urls]
--- a/test_unstructured_inference/inference/test_layout.py
+++ b/test_unstructured_inference/inference/test_layout.py
@@ -273,6 +273,7 @@ def test_from_image_file(monkeypatch, mock_final_layout, filetype, element_extra
         "width": image.width,
         "height": image.height,
         "pdf_rotation": 0,
+        "pdf_rotation_correction": 0,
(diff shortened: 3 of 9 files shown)


vladimir-kivi-ds (approved):
lgtm

